PR #5, opened 2018-07-22
Add sphynx to tagless-optimize

### LINKED ISSUE #2: Move sphynx in

LukaJCB:
Should fix #2 

Will add more docs later :)


kailuowang:
Thanks! I wonder why a separate module. this one has no extra dependencies than core and I would consider it a core function in cats-tagless.

LukaJCB:
You're right, I didn't even think of that, I can switch it back :) though probably still a good idea to have it in a separate package

Issue #6, opened 2018-07-22
PROPOSAL - Migrate Scalameta Macros


diesalbla:
The macro annotations in `cats-tagless` are implemented using the macro system of scalameta. 

However, the development of macros and macro annotations in Scala Meta was closed, as explained 
[here](http://www.scala-lang.org/blog/2017/10/09/scalamacros.html). 

> With these lessons learned, we decided to retire our efforts to build a macro system on top of Scalameta. 

Current support for Scalameta macros may not work in next versions of the scala compiler. 

On the other hand, the Scala compiler has recently [incorporated](https://github.com/scala/scala/pull/6606) the macro-paradise compiler plugin, which enables support for macro annotations based on `scala-reflect`, into their main codebase.

The goal of this issue is to migrate existing macros to the `scala.reflect` based macro annotations.

Odomontois:
For that very reason I temporarily dropped mainecoon from my project, until scalameta paradise didn't get the 2.12.6 support.
This should not happen again!

SemanticBeeng:
@milessabin : in the context of https://www.scala-lang.org/blog/2017/10/09/scalamacros.html is your advise about `scala.reflect` based macros mostly ? 

Any guidance about fit with `scalameta` based macros?
According to the above decision document this way to create macros is adopted.

I understand _" look very hard at non-macro solutions where at all possible"_ but for code generation macros is a good solution and `scalameta` seems to make this productive and sustainable.

Please advise further.

milessabin:
@SemanticBeeng it's about macros in general. The only thing that's certain at this point is that scala.reflect macros are not viable in the long term. It's still unclear what will replace them.

SemanticBeeng:
Understood.
Which leaves me with `scalameta` in mid term, until something better comes along.
Have no knowledge of other better ways to do `program transformation` and `code generation` in Scala.

My best knowledge for long term is Martin's vision here which combines `Scala LMS` and macros (right with me).
`Principled Meta Programming for Scala`
https://gist.github.com/odersky/f91362f6d9c58cc1db53f3f443311140 

Sorry if this extrapolation is "unsafe" but given that `cats-tagless` replaces `mainecoon` this is the best place to mention afaik.

olafurpg:
>  Which leaves me with scalameta in mid term, until something better comes along.

[...]

PR #22, opened 2019-03-06
Blackbox semiauto typeclass derivation

### LINKED ISSUE #8: Semiauto Derivation
It would be very nice to use this with a semiauto encoding in the companion objects. I know circe and shapeless does something like this for Encoder/Decoder/Generic and I'm unsure if it  could be leveraged with the existing macro definitions to create something like the below encoding

```scala
implicit val functorKAlg = deriveFunctorK[MyAlgebra]
```

joroKr21:
As a first step towards migrating cats-tagless to `scala.reflect`
macros, we implement blackbox typeclass derivations macros.

There are many advantages to blackbox macros:
  * strictly less powerful technique than macro annotations
  * entirely type-driven hence cover more corner cases
  * enable semiauto style typeclass derivation for free
  * more friendly to IDEs and other tools (types are known)
  * likely to have more straightforward migration path in Scala 3

The `scala.meta` macro annotations are modified to call the
blackbox macros which means they generate one-liners.

The next step would be to complete replace `scala.meta` with
`scala.reflect` macro annotations.

Ref #6 
Closes #8

### DIFF
--- a/tests/src/test/scala/cats/tagless/tests/autoContravariantTest.scala
+++ b/tests/src/test/scala/cats/tagless/tests/autoContravariantTest.scala
@@ -84,7 +84,7 @@ object autoContravariantTest {
 
   @autoContravariant
   trait AlgWithGenericType[T] {
-    def foo[A](i: T, a: A): Int
+    def foo[U](i: T, a: U): Int
   }
 
   @autoContravariant
--- a/macros/src/main/scala/cats/tagless/Derive.scala
+++ b/macros/src/main/scala/cats/tagless/Derive.scala
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2017 Kailuo Wang
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless
+
+import cats.{Contravariant, FlatMap, Functor, Invariant}
+
+object Derive {
+  def functor[F[_]]: Functor[F] = macro DeriveMacros.functor[F]
+  def contravariant[F[_]]: Contravariant[F] = macro DeriveMacros.contravariant[F]
+  def invariant[F[_]]: Invariant[F] = macro DeriveMacros.invariant[F]
+  def flatMap[F[_]]: FlatMap[F] = macro DeriveMacros.flatMap[F]
+  def functorK[Alg[_[_]]]: FunctorK[Alg] = macro DeriveMacros.functorK[Alg]
+  def invariantK[Alg[_[_]]]: InvariantK[Alg] = macro DeriveMacros.invariantK[Alg]
+  def semigroupalK[Alg[_[_]]]: SemigroupalK[Alg] = macro DeriveMacros.semigroupalK[Alg]
+  def applyK[Alg[_[_]]]: ApplyK[Alg] = macro DeriveMacros.applyK[Alg]
+}
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -1,0 +1,272 @@
+/*
+ * Copyright 2017 Kailuo Wang
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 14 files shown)


kailuowang:
Thanks so much! This looks great to me. As for the fully refined version, I don't know if there are any people actually relying on it (It's not even documented). I'd say let's table that for now.  
The  `mapK` in the companion object is also created to help address refined type issue.

joroKr21:
> how hard would it be to migrate this one?

I suspect it would mean reimplementing the same code (as currently present in scala.meta) with `scala.reflect` macro annotations. But this would mean code duplication with the blackbox macro version. I would rather try to push a bit further and make it work with the macro, because it's purely type-driven. If we define `implicit def fk[A, B]: Derive.functorK[Foo[?[_]] { type T = A; type U = B }]` the macro should just work™

joroKr21:
Now it works with refined types as well.

I also cleaned up the implementation and started factoring out utils but ran out of steam for today. It's looking good though :+1:

joroKr21:
@kailuowang this was easier than I expected. You could take a closer look now if you want, the code is in better shape. The trick of typechecking the declaration first to get enough type information for the implementation seems to work very well.

I tried to factor out as much as possible into reusable functions. Wdyt, would that be enough to derive the other typeclasses? What should we try next? `InvariantK`?

Also it became too tedious to try and provide nice error messages. As it stands if for some reason the conditions for the typeclass are not met, it will just fail to compile with some random error like "type mismatch" or "expected class type, but found mish-mash".

kailuowang:
This code looks way better than what I wrote with Scalameta (sorry about the quasi-quotes spaghetti you had to read). It's also really cool to have refined types working. I think it's in a good place to start tackling other type classes. `InvariantK` should be a great target next.

I would leave the nice error messages for future releases.

joroKr21:
@kailuowang I just pushed an implementation of `InvariantK`.
Also factored out as much as I could into utility functions.
There are a few questions remaining:

* Naming: how should the object holding the semiauto derivations be called (`Derive`, `semiauto` or something else?)
* Should the companion objects of typeclasses defined in cats-tagless also have a `derive` method?

kailuowang:
Nice. 
1. `Derive` sounds better to me. `Derive.functorK[XXX]` sounds better than `semiauto.deriveAutoFunctor[XXX]`
2. Would that complicate the dependency between macros and typeclass definitions? If so, I incline to keep them separate.

joroKr21:
> Would that complicate the dependency between macros and typeclass definitions? If so, I incline to keep them separate.

Ah yes, that might be true. And anyway we can't add `derive` methods to the companion objects of `Functor`, `Apply`, etc. so it would be more consistent to keep all macros in one object.

Update: Now `SemigroupalK` and `ApplyK` also work :sun_with_face: 
Which means it would be easy to implement the lower-kinded versions.

This leaves only `FlatMap`. I'm not sure I understand how it works currently.

kailuowang:
Tha's so cool.  
`FlatMap` was contributed by https://github.com/kailuowang/mainecoon/pull/36 by @Jentsch . It's like implementing `FlatMap` for `Function`, i.e. create a new `Function` as the result of a `flatMap`. but in this case you have to do it for each method in the algebra trait. I can tackle it after this PR if you want.

tests/src/test/scala/cats/tagless/tests/autoContravariantTest.scala
```diff
@@ -84,7 +84,7 @@ object autoContravariantTest {
 
   @autoContravariant
   trait AlgWithGenericType[T] {
-    def foo[A](i: T, a: A): Int
+    def foo[U](i: T, a: U): Int
```
joroKr21: I couldn't get rid of useless type parameter shadowing warnings :disappointed: 
And somehow they become fatal, although I didn't find `-Xfatal-warnings` anywhere.

joroKr21:
@kailuowang this is ready for review! `FlatMap` works as well :tada: 

Just to clarify: for algebra methods that have an implementation we still want to delegate to the underlying algebra, not leave the default implementation, correct? At least that's what the `autoFunctor` tests led me to believe.

Also a small issue: I couldn't get rid of some type parameter shadow warnings. They are meaningless because we reference the types directly but I found no way to suppress them.

tests/src/test/scala/cats/tagless/tests/autoContravariantTest.scala
ChristopherDavenport: Is this a problem with the `A` should the internal reference be some other name? Just trouble shooting as many people use both fatal-warnings and may use a generic type with type named `A`, abstracting internally so it conflicts only with `T_ABSTRACT_TYPE` or something absurdly unlikely to be hit seems possible useful so others don't run into this gotcha.

joroKr21: We never explicitly defined this name. It comes from the method definition itself. I tried to rename it but it's way more complicated than it appears. It's a propagating change because changing the type parameter name leads to changing the method type and method symbol and this change has to be propagated throughout the entire tree.

In the worst case one could use the silencer sbt plugin.

kailuowang:
@joroKr21 I appologize for the delayed response. My last weekend was ... eventful. 

This is so awesome. 

> for algebra methods that have an implementation we still want to delegate to the underlying algebra

That's right. 

The fatal-warning was from `sbt-catalyst` plugin. I think the work around is fine. 

This code looks really neat to me. Let's merge and iterate from here.

dispalt:
whoa amazing nice work @joroKr21

slouc:
+1000 @joroKr21

PR #23, opened 2019-03-12
Blackbox macros: refresh type parameter names

joroKr21:
This is a solution to the issue of type parameter shadowing.
Scalac issues a warning when we define a type parameter `A` if
there is already another one in scope with the same name.
This could be a problem for users with `-Xfatal-warnings` enabled.

The solution is to temporarily set fresh names for all type parameters,
type-check the tree and restore the original names afterwards.
It's a bit hacky, but self-contained.

Reverts type parameters which were renamed to work around the issue.

Follow up #22 
cc @ChristopherDavenport

### DIFF
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -25,6 +25,7 @@ class DeriveMacros(val c: blackbox.Context) {
   import c.internal._
   import c.universe._
 
+  /** A reified method definition with some useful methods for transforming it. */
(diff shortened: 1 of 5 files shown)


joroKr21:
On no, it still issues the warning, although I can see in the tree that all type parameters were replaced. Maybe that's the warning that originated with our own call to `c.typecheck`?

joroKr21:
> Maybe that's the warning that originated with our own call to c.typecheck?

Yep, that's true. The warnings are emitted even with `silent = true`

joroKr21:
There, fixed it with a hack :smile: Ideally we should be able to suppress warnings when we typecheck inside a macro but there is no such API.

PR #26, opened 2019-03-19
extract the derive to a new module / renamed old one to legacy-macro

kailuowang:
As an intermediate step for our migration plan, we renamed the old module to `legacy-macro` and make the new one `macro`. 

I also added doctests to sanity check the newly extracted module. Once this is merged, we can release it as 0.3.0


kailuowang:
@joroKr21 Thanks. I propose we add new macro annotations to this new `macro` module. Not sure if we need another module.

PR #28, opened 2019-03-21
added scala 2.13 to build

kailuowang:
Note that's we are only running the tiny scaladoc tests on 2.13 for `cats-tagless-macros`. So the test coverage on 2.13 is next to zero. But I think that's okay because 2.13 is still on milestone release, we treat this as a milestone release as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,7 @@ addCommandAlias("gitSnapshots", ";set version in ThisBuild := git.gitDescribedVe
 
 addCommandAlias("validateJVM", ";testsJVM/test ; docs/makeMicrosite")
 
-lazy val libs = org.typelevel.libraries
+lazy val libs = org.typelevel.libraries.add("kind-projector", "0.9.9")
 
 val apache2 = "Apache-2.0" -> url("https://www.apache.org/licenses/LICENSE-2.0.html")
 val gh = GitHubSettings(org = "typelevel", proj = "cats-tagless", publishOrg = "org.typelevel", license = apache2)
@@ -137,7 +137,6 @@ lazy val docs = project
     scalacOptions in Tut ~= (_.filterNot(Set("-Ywarn-unused-import", "-Ywarn-dead-code"))),
     git.remoteRepo := gh.repo,
     includeFilter in makeSite := "*.html" | "*.css" | "*.png" | "*.jpg" | "*.gif" | "*.js" | "*.swf" | "*.yml" | "*.md")
-
 
 
 lazy val buildSettings = sharedBuildSettings(gh, libs)
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,9 @@ lazy val lawsM   = module("laws", CrossType.Pure)
   .enablePlugins(AutomateHeaderPlugin)
 
 
-lazy val legacyMacros    = prj(legacyMacrosM)
+lazy val legacyMacros    = prj(legacyMacrosM).settings(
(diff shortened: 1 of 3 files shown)


kailuowang:
This should allow us to release on scala 2.13-M5

project/plugins.sbt
```diff
@@ -1,5 +1,5 @@
-addSbtPlugin("org.typelevel" %% "sbt-catalysts" % "0.11.3")
+addSbtPlugin("org.typelevel" %% "sbt-catalysts" % "0.12")
```
joroKr21: Thank you for updating catalyst :heart:

build.sbt
```diff
@@ -173,6 +175,15 @@ lazy val metaMacroSettings: Seq[Def.Setting[_]] = Seq(
   sources in (Compile, doc) := Nil // macroparadise doesn't work with scaladoc yet.
 )
 
+def scalatestVersion(scalaVersion: String): String =
+  if (priorTo2_13(scalaVersion)) "3.0.5" else "3.0.6-SNAP5"
```
joroKr21: Two questions:
  1. Why does it depend on the Scala version?
  2. Why a snapshot? There is already scalatest 3.0.7

build.sbt
```diff
@@ -42,7 +42,7 @@ lazy val rootJS = project
   )
 
 
-lazy val core    = prj(coreM)
+lazy val core    = prj(coreM).settings(scala213Setting)
```
joroKr21: Shouldn't this be taken care of by the `commonSettings`?

kailuowang: Common settings are shared by all modules but we can't include this in legacy macros module, tests and no need to include it in laws module (which is a bit tricky due to scalacheck and cats dependency )

build.sbt
kailuowang: I made it way due to the concern of compatibility with cats - cats' testkit depends on these libraries this way, i.e different version on different Scala versions (because it needs to maintain maximum backwards BC).
To be clear macros module does not depends on cats testkit yet but it might soon. It will be simplified once cats release on Scalatest 3.1.0 and scalacheck 1.14

build.sbt
joroKr21: But `commonSettings` include Scala 2.13 which is also exactly what `scala213Settings` do"

`crossScalaVersions := Seq(libs.vers("scalac_2.11"), scalaVersion.value, libs.vers("scalac_2.13"))`

Maybe I'm missing something.

build.sbt
```diff
@@ -137,13 +140,12 @@ lazy val docs = project
...
 
 
-
 lazy val buildSettings = sharedBuildSettings(gh, libs)
 
 lazy val commonSettings = sharedCommonSettings ++ Seq(
   parallelExecution in Test := false,
   scalaVersion := libs.vers("scalac_2.12"),
-  crossScalaVersions := Seq(libs.vers("scalac_2.11"), scalaVersion.value),
+  crossScalaVersions := Seq(libs.vers("scalac_2.11"), scalaVersion.value, libs.vers("scalac_2.13")),
```
joroKr21: @kailuowang these are the settings I meant concerning `scala213Settings`

build.sbt
joroKr21: Ok, that explains it :+1:

build.sbt
kailuowang: Oops. Missed that

kailuowang:
Thanks I am thinking a new version since this one bumped some dependences across

PR #30, opened 2019-03-22
build changes to fix release on 2.13

kailuowang:
First attempt to release on scala 2.13 failed. These changes should fix it.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -5,4 +5,3 @@
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
 -XX:+CMSClassUnloadingEnabled
--XX:+UseConcMarkSweepGC
--- a/.jvmopts
+++ b/.jvmopts
@@ -3,5 +3,3 @@
 -Xmx3G
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
--XX:-UseGCOverheadLimit
--XX:+CMSClassUnloadingEnabled
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,8 @@
+-Dfile.encoding=UTF8
+-Xms1G
+-Xmx3G
+-XX:ReservedCodeCacheSize=250M
+-XX:+TieredCompilation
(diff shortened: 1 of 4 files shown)


.jvmopts
```diff
@@ -0,0 +1,8 @@
+-Dfile.encoding=UTF8
+-Xms1G
+-Xmx3G
+-XX:ReservedCodeCacheSize=250M
+-XX:+TieredCompilation
+-XX:-UseGCOverheadLimit
+-XX:+CMSClassUnloadingEnabled
+-XX:+UseConcMarkSweepGC
```
joroKr21: Unfortunately these GC options would prevent loading the project locally if you set a different GC in `$SBT_OPTS` because they just get appended and Java doesn't tolerate conflicting GC options. 

Anyway the default GC in Java 8 is G1GC which can also unload classes. So can we just remove the GC related options?

build.sbt
```diff
@@ -146,8 +149,16 @@ lazy val commonSettings = sharedCommonSettings ++ Seq(
...
   crossScalaVersions := Seq(libs.vers("scalac_2.11"), scalaVersion.value),
-  developers := List(Developer("Kailuo Wang", "@kailuowang", "kailuo.wang@gmail.com", new java.net.URL("http://kailuowang.com")))
-) ++ scalacAllSettings ++ unidocCommonSettings ++
+  //todo: re-enable disable scaladoc on 2.13 due to https://github.com/scala/bug/issues/11045
+  sources in (Compile, doc) := {
+    val docSource = (sources in (Compile, doc)).value
+    if (priorTo2_13(scalaVersion.value)) docSource else Nil
+  },
+  developers := List(
+    Developer("Georgi Krastev", "@joroKr21", "joro.kr.21@gmail.com", new java.net.URL("https://www.linkedin.com/in/georgykr")),
```
joroKr21: Yikes I have to update my LinkedIn now :smile: or start a blog...

.jvmopts
kailuowang: No problem, removed.

PR #33, opened 2019-03-25
Try to override methods more conservatively

### LINKED ISSUE #31: Typeclass derivation should not try to override final methods
This is obviously a bug but also an enhancement because once it's fixed we could define final derived methods on algebras which would not be subject to the restrictions for deriving the typeclass.

### LINKED ISSUE #32: Default arguments should be supported
Compilation will fail, if the algebra has default arguments.
As an example, the following code will not compile
```
import cats._, cats.implicits._, cats.data._
import cats.tagless._, cats.tagless.implicits._


trait Alg[F[_]] {
  def test(msg: String = "Hello World"): F[String]
}

object Alg {
  implicit def functorK: FunctorK[Alg] = Derive.functorK[Alg]
}

class AlgInterpreter extends Alg[Id] {
  def test(msg: String) = msg
}

implicit val idToOption = new (Id ~> Option) {
  override def apply[A](a: Id[A]): Option[A] = Some{ a }
}

val idAlg = new AlgInterpreter
val optionAlg = idAlg.mapK(idToOption)

println(idAlg.test())
println(optionAlg.test())
```
When the default value of the msg argument of the test function is removed, the compilation will succeed.

joroKr21:
Do not try to override:
  * final methods
  * bridges / artifacts
  * synthetic methods (this covers default parameters)

fixes #31 
fixes #32 

For now I just added some `autoFunctorK` tests. We should decide whether to write separate tests for the macros or keep piggybacking on the macro annotation tests.

### DIFF
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -30,16 +30,17 @@ class DeriveMacros(val c: blackbox.Context) {
     def typeArgs: List[Type] = for (tp <- tps) yield typeRef(NoPrefix, tp.symbol, Nil)
     def paramLists(f: Type => Type): List[List[ValDef]] = for (ps <- pss)
       yield for (p <- ps) yield ValDef(p.mods, p.name, TypeTree(f(p.tpt.tpe)), p.rhs)
-    def argLists(f: TermSymbol => Tree): List[List[Tree]] = for (ps <- pss)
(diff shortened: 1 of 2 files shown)


joroKr21:
Yikes I think trait encoding changed in 2.12. What now?

joroKr21:
> Yikes I think trait encoding changed in 2.12. What now?

Phew, it was not the trait encoding, it was the default parameter flag :sweat_smile:

PR #42, opened 2019-05-06
fixed laws compilation on 2.13 and include it in build

kailuowang:


laws/src/main/scala/cats/tagless/laws/discipline/ApplyKTests.scala
```diff
@@ -33,15 +33,11 @@ trait ApplyKTests[F[_[_]]] extends FunctorKTests[F] with SemigroupalKTests[F] {
                                                ArbCG: Arbitrary[F[B]],
                                                ArbCH: Arbitrary[F[C]],
                                                iso: IsomorphismsK[F],
-                                               ArbitraryG: Arbitrary[A[T]],
-                                               ArbitraryH: Arbitrary[B[T]],
-                                               ArbitraryI: Arbitrary[C[T]],
```
kailuowang: turns out that these are never used

Issue #37, opened 2019-04-29
replace scala-meta macro annotations with scalamacros


kailuowang:
This migration involves three areas of logic, in priority high to low order.  We can do these in separate PRs. 

- [x] derivation instances for algebra without abstract type members - one that generates default instances using the `cats.tagless.Derive` . e.g. 
```scala
  implicit val functorKForAlg: FunctorK[Alg] =
      Derive.functorK[Alg]
```
This need to handle situation where Alg with extra type params other than the effect type
The actual derivation is provide by the `cats.tagless.Derive` from the cats-tagless-macro module, the annotation just need to generate delegate methods (main challenge is to deal with different shapes of Alg types).

- [ ] derivation instances for algebra with abstract type members. We need normal instances and instances that are fully refined under an object in the companion named `fullyRefined` , see https://github.com/typelevel/cats-tagless/blob/master/legacy-macros/src/main/scala/cats/tagless/autoFunctorK.scala#L72
for existing scala meta implementation. again the actual derivation is supported by the `cats.tagless.Derive` from the cats-tagless-macro module
 
- [x] optional, alg instance summoning method like `def apply[F[_]](implicit ev: Alg[F]): Alg[F]`

marcin-rzeznicki:
Hey, thanks for the issue. I am given some paid time from my company to participate in Sustainability Program, so I thought I could give it a try - I obviously don't want to step in anyone's toes, so please let me know if someone is already working on this, but if not I think I could take care of it.

kailuowang:
cc @joroKr21

dispalt:
I sorta half started working on it here, https://github.com/dispalt/tagless-redux/blob/master/macros/src/main/scala/cats/tagless/autoFunctorK.scala but it is far from complete, and the code is really messy.  But if you wanted to look at something there you go.

I am not going to work on it anytime soon (unfortunately).

joroKr21:
I am not working on this actively atm. I would be very happy if you can make this happen. Even if you get only part of the way,  don't be shy to open a PR and I will do my best to help.

All I did was remove some code which became unused after delegating companion methods to the generated instances. Just to reduce the surface area. I will push it later tonight.

joroKr21:
> All I did was remove some code which became unused after delegating companion methods to the generated instances. Just to reduce the surface area. I will push it later tonight.

Here it is: #38

marcin-rzeznicki:
Thanks guys, I'll take a stab at it

[...]

PR #43, opened 2019-05-06
[WIP] Scala meta out

marcin-rzeznicki:
Hey,
I am making this PR to show the progress of migration from scala-meta annotations. I was able to replace  `@autoFunctor` and `@autoFlatMap`. Replacement of remaining annotations will be very similar I assume, so I think this is good moment to comment on the code - if you see anything weird, questionable, please do not hesitate to comment. 
Tests are passing!
PS. After all of the annotations are migrated, I will have to modify the build to get rid of scala-meta and pull in macro-paradise explicitly
PPS. There is no need to merge this PR - once I am done, I will prepare the final PR with all the changes

UPDATE: I've added `@autoInvariant` and `@autoContravariant` as well - they're all roughly the same


kailuowang:
Thanks!!! I can't say I am scala-reflect macro expert, but this looks sane to me. @joroKr21 WDYT?

joroKr21:
Looks good so far :+1: 

Great stuff @marcin-rzeznicki ! We can review details later, but I like the approach.

I have some a question in general - how much of the old macros do we want to port? Companion object methods and fully refined instances? `apply` method? These are also things we can leave for future work.

marcin-rzeznicki:
Thanks! I am migrating "higher-kinded" annotations then

kailuowang:
@joroKr21 to answer
> how much of the old macros do we want to port? Companion object methods and fully refined instances? apply method? These are also things we can leave for future work.

I tried to spec out requirements in #37. I agree they can be done in multiple stages / PRs.

marcin-rzeznicki:
Actually, after looking more deeply at the code, I think that #37 can be done in one sweep easily. Handling of type members is not that complicated and it's already done for scala meta, there are existing tests etc.

marcin-rzeznicki:
Closing this pull request

PR #44, opened 2019-05-10
Fix DeriveMacros when a higher-kinded refined type is passed as a ref

joroKr21:
One can pass a refined type `Hk[F[_]] { type T }` as a reference:

```scala
Derive.functorK[Hk { type T = Int }]
```

and this currently breaks the derivation.

To resolve the problem we `etaExpand` the type constructor to
obtain a `PolyType` which is easier to work with.

As a cleanup, extract the initial normalization performed on
the algebra type parameter to a helper method.

Discovered by @marcin-rzeznicki as part of #37


marcin-rzeznicki:
Ok, so I actually reverted the problematic behavior, but if this fixes HK stuff then I'll rebase on it. Great work! Thanks :+1:

kailuowang:
This is now included in #45

joroKr21:
Merged with #45

Issue #48, opened 2019-06-01
Profunctor/Bifunctor derivation


joroKr21:
Adding an `@autoProfunctor` annotation would require reworking `AlgDefn` a bit to handle a list of type parameters instead of only one.

LukaJCB:
We could probably derive bifunctor in a similar way :)

marcin-rzeznicki:
Good idea! I can try to add that some time next week, now that I have quite recent understanding of the code base

joroKr21:
> We could probably derive bifunctor in a similar way :)

Indeed. Bifunctors are really popular these days :smile: 

@marcin-rzeznicki I have an old branch that adds `Derive.profunctor`, but not the annotation.
I will add some tests and make a PR soon.

milessabin:
> We could probably derive bifunctor in a similar way :)

For reference: https://github.com/dotty-staging/dotty/blob/4e434e0f7a6b0f62dc39db19c3b65139723d7548/tests/typeclass-derivation/typeclass-derivation-4-split/TypeClasses_3.scala#L136-L181.

marcin-rzeznicki:
Thanks @milessabin . Also thanks for the PR @joroKr21

marcin-rzeznicki:
BTW we could change the tile and mention `Bifunctor` /

marcin-rzeznicki:
Sorry for the long delay. I was attending a Haskell conference for about a week plus had things to do in my work :smile:  I'm on it now - will PR soon.

marcin-rzeznicki:
I think we can close this. Thanks @kailuowang @joroKr21

joroKr21:
Resolved in #57, thanks @marcin-rzeznicki!

PR #50, opened 2019-06-03
Add Profunctor derivation macro

joroKr21:
Better to add the macro annotation in another PR,
because `MacroUtils` would have to be refactored
to handle algebras with multiple effect type parameters.

Ref #48

### DIFF
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -16,6 +16,7 @@
 
 package cats.tagless
 
+import cats.arrow.Profunctor
 import cats.data.Tuple2K
 import cats.{Contravariant, FlatMap, Functor, Invariant}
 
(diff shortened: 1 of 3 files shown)


macros/src/main/scala/cats/tagless/DeriveMacros.scala
```diff
@@ -148,6 +149,8 @@ class DeriveMacros(val c: blackbox.Context) {
     typeCheckWithFreshTypeParams(q"{ $definition; new ${declaration.symbol} }")
   }
 
+  // def map[A, B](fa: F[A])(f: A => B): F[B]
```
kailuowang: nit,  the `def mapK` comment is helpful, but is the `def map` comment needed here?

macros/src/main/scala/cats/tagless/DeriveMacros.scala
```diff
@@ -177,6 +181,8 @@ class DeriveMacros(val c: blackbox.Context) {
       implement(appliedType(algebra, g.asType.toTypeConstructor), types ++ methods)
     }
 
+  // def imap[A, B](fa: F[A])(f: A => B)(g: B => A): F[B]
```
kailuowang: again, is the `def imap` comment here needed?

macros/src/main/scala/cats/tagless/DeriveMacros.scala
joroKr21: I'm not sure. The thing is that the macro implementation is generic enough to cover both. I don't know what's the best way to express this (or indeed if we even want to express this). Should I remove the comment?

marcin-rzeznicki:
Thank you, this will be immensely helpful for the annotation - I'll also add derivation for `bimap` in the following way. LGTM (if that matters) :+1:

macros/src/main/scala/cats/tagless/DeriveMacros.scala
joroKr21: I will merge it like this for now. We can remove the comment later, it's not a problem.
But I still think it makes some sense to clarify that it covers both methods.

PR #54, opened 2019-06-12
Add support for repeated (vararg) parameters

### LINKED ISSUE #53: Support varargs
Currently (0.7) this fails to derive:

```scala
@autoFunctorK
trait Foo[F[_]] {
  def bar(is: Int*): F[Int]
}
```

joroKr21:
They require special handling because they are modelled as a
magic type constructor `<repeated>[T]` in the compiler.

Fixes #53 

I can add compile tests about other typeclasses later.
I'm not sure if this needs runtime tests?


joroKr21:
Just added some more tests so I'm merging this.

PR #55, opened 2019-06-17
update to latest scala 2.13

kailuowang:
Turned out to be more involved due to Cats' change on test API regarding Function equality.

### DIFF
--- a/tests/src/test/scala/cats/tagless/tests/CatsTaglessTestSuite.scala
+++ b/tests/src/test/scala/cats/tagless/tests/CatsTaglessTestSuite.scala
@@ -19,10 +19,12 @@ import cats.tests.CatsSuite
 import cats.arrow.FunctionK
 import cats.instances.AllInstances
 import cats.kernel.Eq
+import cats.laws.discipline.ExhaustiveCheck
+import cats.laws.discipline.ExhaustiveCheck.instance
 import cats.tagless.syntax.AllSyntax
 import org.scalacheck.{Arbitrary, Gen}
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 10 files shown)


tests/src/test/scala/cats/tagless/tests/CatsTaglessTestSuite.scala
```diff
@@ -36,4 +38,29 @@ trait TestInstances {
...
 
   implicit val eqThrow: Eq[Throwable] = Eq.allEqual
+
+  implicit def exhaustiveCheckForTry[A](implicit A: ExhaustiveCheck[A]): ExhaustiveCheck[Try[A]] =
+    instance(Stream.cons(Failure(new RuntimeException("some exception")), A.allValues.map(Success(_))))
+
+  implicit def exhaustiveCheckForList[A](implicit A: ExhaustiveCheck[A]): ExhaustiveCheck[List[A]] =
+    instance(Stream.cons(Nil, A.allValues.map(List(_))))
+
+  implicit def exhaustiveCheckForString: ExhaustiveCheck[String] = new ExhaustiveCheck[String] {
```
kailuowang: These instances are very naive but probably good enough for tagless alg transformation tests in which the function implementation are randomly generated as well.

kailuowang:
@LukaJCB this turned out to be more involved, due to breaking change in Cats test API. mind to take another look?

tests/src/test/scala/cats/tagless/tests/CatsTaglessTestSuite.scala
joroKr21: We could randomize them if we wanted. But honestly I don't think it's necessary.

PR #56, opened 2019-06-23
Flexible derivation based on existing instances + Instances for cats.data

### LINKED ISSUE #9: Provide instances for cats core types
e.g. 
`FunctorK` for `EitherT`, `StateT`, `WriterT`,`Kleisli`
`InvariantK` for `Functor`, `Apply`, `FlatMap`, etc...
I wonder if they are welcomed in the companion or separate module?

### LINKED ISSUE #10: autoFunctorK should respect instances of result types
e.g.
```scala
@autoFunctorK
trait Foo[F[_]]{
   def int: F[Int]
}

//currently doesn't work
@autoFunctorK
trait Baz[F[_]]{
  def foo: Foo[F]
}
```

### LINKED ISSUE #52: autoInvariantK fails when function takes lambda
`@autoInvariantK` fails on this code

```scala
@autoInvariantK
trait InvariantFail[F[_]] {

  def function(f: String => F[Boolean]): F[Boolean]
}
```

Error
```
[error] object creation impossible, since method function in trait InvariantFail of type (f: String => G[Boolean])G[Boolean] is not defined
[error] method function overrides nothing.
[error] Note: the super classes of <$anon: ore.auth.InvariantFail[G]> contain the following, non final members named function:
[error] def function(f: String => G[Boolean]): G[Boolean]
[error] object creation impossible, since method function in trait InvariantFail of type (f: String => G[Boolean])G[Boolean] is not defined
[error] method function overrides nothing.
[error] Note: the super classes of <$anon: ore.auth.InvariantFail[G]> contain the following, non final members named function:
[error] def function(f: String => G[Boolean]): G[Boolean]
[error] four errors found
```

joroKr21:
I was too excited to wait for completing this PR. It's going way smoother than I expected :astonished: 

Derivation based on existing typeclass instances can work for more complex parameter and return types, such as `EitherT`, `List`, `fs2.Stream` and even builder style algebras.

It looks like we only need the `Id1` type alias, not even `Const1`.

TODO:
 - [x] Add instances for `cats.data` types
 - [x] Tests
 - [x] Cleanup

Where should we put instances? In the companion objects or in a `cats.tagless.instances` package?

closes #9 
closes #10 
closes #52

### DIFF
--- a/core/src/main/scala/cats/tagless/instances/package.scala
+++ b/core/src/main/scala/cats/tagless/instances/package.scala
@@ -16,4 +16,17 @@
 
 package cats.tagless
 
-package object implicits extends instances.AllInstances with syntax.AllSyntax
+package object instances {
+  object all extends AllInstances
+  object eitherK extends EitherKInstances
+  object eitherT extends EitherTInstances
+  object func extends FuncInstances
+  object idT extends IdTInstances
+  object iorT extends IorTInstances
+  object kleisli extends KleisliInstances
+  object nested extends NestedInstances
+  object oneAnd extends OneAndInstances
+  object optionT extends OptionTInstances
+  object tuple2K extends Tuple2KInstances
+  object writerT extends WriterTInstances
+}
--- a/core/src/main/scala/cats/tagless/instances/AllInstances.scala
+++ b/core/src/main/scala/cats/tagless/instances/AllInstances.scala
@@ -14,6 +14,16 @@
  * limitations under the License.
  */
 
-package cats.tagless
+package cats.tagless.instances
 
-package object implicits extends instances.AllInstances with syntax.AllSyntax
+trait AllInstances extends EitherKInstances
+  with EitherTInstances
+  with FuncInstances
+  with IdTInstances
+  with IorTInstances
+  with KleisliInstances
+  with NestedInstances
+  with OneAndInstances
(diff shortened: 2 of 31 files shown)


kailuowang:
Wow sweet! Thanks!
For instances, I vote for placing them in cats.tagless.instances

joroKr21:
This turned into a huge PR but is ready for review now.

core/src/main/scala/cats/tagless/instances/package.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package cats.tagless
+
+package object instances {
+  object all extends AllInstances
+  object eitherK extends EitherKInstances
+  object eitherT extends EitherTInstances
+  object func extends FuncInstances
+  object idT extends IdTInstances
+  object iOrT extends IorTInstances
```
catostrophe: Strange naming

joroKr21: I agree but I think `IorT` is the culprit. Do you think we should name it `iorT`?

joroKr21: I changed it to `iorT` because I agree `iOrT` looks too weird. We can revise this decision later.

PR #57, opened 2019-06-24
autoProfunctor / bifunctor

marcin-rzeznicki:
Adds `@autoProfunctor` annotation (see #48 ).

I needed to add more elaborate handling of algebra "shapes". Before profunctor the code had had to deal with rather limited set of algebras - `Alg[F[_], ...]` or `Alg[F,..]`, while to support profunctor (and, later, bifunctor) one had to consider `Alg[F1, F2]` . I decided to go with rather elaborate machinery of `AlgebraResolver` which produces metadata about the algebra being processed based on the expected position and kind of effect type params eg. `AnyLastTypeParam` or `FirstHigherKindedTypeParam`. The expected type of algebra (currently either `UnaryAlgebra` or `BinaryAlgebra`) is carried as its type member and used as a dependent type in `enrichAlgebra`. While this might be a cognitive overkill, it has some nice properties compared to reimplementing algebra definition to use `Seq` or something similar:
- you cannot use something like `newTypeSig(..)` with wrong number of params
- you can catch errors earlier eg. the difference between `@finalAlg` and `@autoFunctor`
- it has the potential for better error reporting (see #52 )

Please review

### DIFF
--- a/macros/src/main/scala/cats/tagless/Derive.scala
+++ b/macros/src/main/scala/cats/tagless/Derive.scala
@@ -17,7 +17,7 @@
 package cats.tagless
 
 import cats.arrow.Profunctor
-import cats.{Contravariant, FlatMap, Functor, Invariant}
+import cats.{Bifunctor, Contravariant, FlatMap, Functor, Invariant}
 
 object Derive {
 
@@ -25,6 +25,7 @@ object Derive {
   def contravariant[F[_]]: Contravariant[F] = macro DeriveMacros.contravariant[F]
   def invariant[F[_]]: Invariant[F] = macro DeriveMacros.invariant[F]
   def profunctor[F[_, _]]: Profunctor[F] = macro DeriveMacros.profunctor[F]
+  def bifunctor[F[_, _]]: Bifunctor[F] = macro DeriveMacros.bifunctor[F]
   def flatMap[F[_]]: FlatMap[F] = macro DeriveMacros.flatMap[F]
 
   /**
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -18,7 +18,7 @@ package cats.tagless
 
 import cats.arrow.Profunctor
 import cats.data.Tuple2K
-import cats.{Contravariant, FlatMap, Functor, Invariant}
+import cats.{Bifunctor, Contravariant, FlatMap, Functor, Invariant}
 
 import scala.reflect.macros.blackbox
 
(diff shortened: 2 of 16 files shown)


marcin-rzeznicki:
Will we ever need _fullyRefined_ for this :question:

marcin-rzeznicki:
I also added a `Bifunctor` derivation plus the usual annotation. I am not sure if my macro handles all the cases correctly, so please double-check (@joroKr21  ping)

joroKr21:
`Bifunctor` macro LGTM, I will check the rest later :+1:

marcin-rzeznicki:
Thank you so much!

marcin-rzeznicki:
Thanks a lot for the review. 
 > I think I see why you needed to distinguish between UnaryAlgebra and BinaryAlgebra and couldn't just use a list of type parameters, it's because of companion methods and auto-derive isn't it? 

Yes, exactly. That's where you need to be aware of how many type args you actually need to reify the algebra type. I was also thinking about making algebra _shapes_ even more specialized by taking the kind of the type param into account, but eventually I thought it'd become too complex.

As for the _fullyRefined_, `autoFunctor` also does not emit it. For now I can remove the possibility to do that from `BinaryAlgebra` (that's the additional flexibility we gain from this model - that some operations can be defined only for algebra shapes that actually support them). It will bump up the coverage a bit :smile: Just let me know what you think :-)

joroKr21:
I think it's fine to keep the fully refined helpers. I'm thinking is there actually any disadvantage to always generating the fully refined signature? Even if there is a good reason not to, we have at least two other sensible options:
  1. Add a `fullyRefined` flag to macro annotations
  2. Only generate when there are actual type members to refine

I think we're still good on coverage > 90% that's excellent. Although I'm not sure how it's measured :smile:

joroKr21:
@kailuowang @LukaJCB Do we always squash and merge? I think the commits here are usefully split, but we also don't have rebase and merge enabled.

kailuowang:
@joroKr21, I just enabled it.

marcin-rzeznicki:
@joroKr21 I don't think there is any disadvantage (except for maybe slightly longer compilation time), but I must admit that I'm not really sure what are the use cases for `autoRefine`. Still, if I were to decide, I'd create a flag as you proposed (I can do it). This way the user knows that import is always available whenever this flag is set. I think that if we generated only when it is needed, then code would just stop compiling after eg. removing the last type member which can be infuriating, especially when there is no good tooling support for macros - so you can't rely on, say, IntelliJ "squiggles" because everything is red anyway :smile: But, I am happy with any outcome

joroKr21:
I think we need a separate issue for `autoRefine`

kailuowang:
Thanks @marcin-rzeznicki.
Test coverage should be legit. The refined feature is indeed advanced. It definitely adds work. I think a flag in the annotation is a solution. I think we don't have to always provide this feature for all derivations. I don't want it to be an unnecessary burden when adding new derivations. A flag will make it clear to users which derivations have this feature and which don't.

Issue #51, opened 2019-06-08
Semigroupal derivation


joroKr21:
See `SemigroupalK` and its derivation.

marcin-rzeznicki:
Got it, I still haven't got time to look at Bifunctor and friends but I'm optimistic about it :-)

PR #95, opened 2019-12-12
Instrument

Fristi:
As discussed with @joroKr21 on gitter.

It would be nice to have a typeclass `DescribeK` which includes the method name (and class name) for all the algebra methods plus the resulting computation. This can be used by tracing or metrics.

Pseudo example code:

```scala
trait OffsetStore[F[_]] {
  def get(name: String): F[Option[Long]]
  def set(name: String, eventId: Long): F[Int]
}

val dummy = new OffsetStore[RIO[Tracing, *]] {
  def get(name: String): RIO[Tracing, Option[Long]] = ZIO.succeed(Some(2l))
  def set(name: String, eventId: Long): RIO[Tracing, [Int] = ZIO.succeed(1)
}

val describeK = Derive.describe[OffsetStore]
val functorK = Derive.functorK[OffsetStore]

// transform the describeK algebra using a natural transformation which uses `RIO[Tracing, *]` and the method `String` 
// to create a OpenTracing span by using `.instrumented`
val tracing = new (Lambda[A => (RIO[Tracing, A], String)] ~> RIO[Tracing, *]) {
    def apply[A](fa: (RIO[Tracing, A], String)): RIO[Tracing, A] =
      fa._1.instrumented(fa._2)
  }

//instrumented algebra
val instrumented = functorK.mapK(describeK)(tracing)
```

The algebra plus macro is there, but the only thing I'm stuck at is how to create a type lambda in macro code? Maybe someone could jump in and help :) 

See:

```
Error:(14, 31) exception during macro expansion: 
scala.reflect.macros.TypecheckException: type mismatch;
 found   : OffsetStore[F$macro$2]
 required: OffsetStore[[A](F$macro$2[A], String)]
```

As you can see, the required type does match up with the given (which was `f`)

Also we need some tests.

### DIFF
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -18,6 +18,7 @@ package cats.tagless
 
 import cats.arrow.Profunctor
 import cats.data.Tuple2K
+import cats.tagless.diagnosis.{Instrument, Instrumentation}
 import cats.{Bifunctor, Contravariant, FlatMap, Functor, Invariant, Semigroupal}
 
 import scala.reflect.macros.blackbox
@@ -540,6 +541,30 @@ class DeriveMacros(val c: blackbox.Context) {
       implement(algebra)(c, d)(types ++ methods)
     }
 
+  // def instrument[F[_]](af: Alg[F]): Alg[Instrumentation[F, *]]
+  def instrumentation(algebra: Type): (String, Type => Tree) =
+    "instrument" -> { case PolyType(List(f), MethodType(List(af), _)) =>
+      val Instrumentation = symbolOf[Instrumentation[Any, Any]]
+      val F = f.asType.toTypeConstructor
+      val Af = singleType(NoPrefix, af)
+      val members = overridableMembersOf(Af)
+      val types = delegateAbstractTypes(Af, members, Af)
+      val algebraName = algebra.typeSymbol.name.decodedName.toString
+
+      val methods = delegateMethods(Af, members, af) {
+        case method if method.returnType.typeSymbol == f =>
+          val body = q"_root_.cats.tagless.diagnosis.Instrumentation(${method.body}, $algebraName, ${method.name.decodedName.toString})"
+          val returnType = appliedType(Instrumentation, F :: method.returnType.typeArgs)
+          method.copy(body = body, returnType = returnType)
+        case method if method.occursInSignature(f) =>
+          abort(s"Type parameter $f can only appear as a top level return type in method ${method.name}")
+      }
+
+      val instrumentationType = appliedType(Instrumentation, F :: F.typeParams.map(_.asType.toType))
+      val DescribedAlg = appliedType(algebra, polyType(F.typeParams, instrumentationType))
+      implement(DescribedAlg)()(types ++ methods)
+    }
+
   def functor[F[_]](implicit tag: WeakTypeTag[F[Any]]): Tree =
     instantiate[Functor[F]](tag)(map)
 
@@ -578,4 +603,7 @@ class DeriveMacros(val c: blackbox.Context) {
 
   def applyK[Alg[_[_]]](implicit tag: WeakTypeTag[Alg[Any]]): Tree =
     instantiate[ApplyK[Alg]](tag)(mapK, productK)
+
+  def instrument[Alg[_[_]]](implicit tag: WeakTypeTag[Alg[Any]]): Tree =
+    instantiate[Instrument[Alg]](tag)(instrumentation)
 }
--- a/macros/src/main/scala/cats/tagless/autoInstrument.scala
+++ b/macros/src/main/scala/cats/tagless/autoInstrument.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2019 cats-tagless maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless
+
+import scala.annotation.{StaticAnnotation, compileTimeOnly}
+import scala.collection.immutable.Seq
+import scala.reflect.macros.whitebox
+
+/** Auto generates an instance of `cats.Invariant`. */
+@compileTimeOnly("Cannot expand @autoInstrument")
+class autoInstrument extends StaticAnnotation {
+  def macroTransform(annottees: Any*): Any = macro autoInstrumentMacros.instrumentInst
+}
+
+private[tagless] class autoInstrumentMacros(override val c: whitebox.Context) extends MacroUtils  {
+  import c.universe._
+
+  private def generateInstrumentFor(algebraName: String)(algebraType: Tree, typeParams: Seq[TypeDef]) =
+    typeClassInstance(
+      TermName("instrumentFor" + algebraName),
+      typeParams,
+      tq"_root_.cats.tagless.diagnosis.Instrument[$algebraType]",
+      q"_root_.cats.tagless.Derive.instrument[$algebraType]"
+    )
+
+  def instrumentInst(annottees: c.Tree*): c.Tree =
+    enrichAlgebra(annottees.toList, AlgebraResolver.FirstHigherKindedTypeParam) { algebra =>
+      algebra.forVaryingEffectType(generateInstrumentFor(algebra.name)) :: Nil
+    }
+}
--- a/tests/src/test/scala/cats/tagless/tests/InstrumentTests.scala
+++ b/tests/src/test/scala/cats/tagless/tests/InstrumentTests.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 cats-tagless maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


kailuowang:
Thanks I like the idea. Not 100% sure about the TC name `DescribeK` , Maybe `Instrument`? not sure if we need a `K` there. 
In the meantime you can find some macro code working with type lambda here .
https://github.com/typelevel/cats-tagless/blob/c8b0e71fe33b29341df28b5cf6276747fe4a9af1/macros/src/main/scala/cats/tagless/MacroUtils.scala#L140

joroKr21:
This is still on the level of `DeriveMacros` so you can take a look at `productK` for inspiration.
But instead of `a => Tuple2K[F, G, a]` we want to have `a => (F[a], String)`.

I'm also still not sure if it would be useful to have just the `a => String` version for anything.

macros/src/main/scala/cats/tagless/DeriveMacros.scala
```diff
@@ -490,6 +490,26 @@ class DeriveMacros(val c: blackbox.Context) {
...
 
+  // def describe[F[_]](af: Alg[F]): Alg[λ[a => (F[a], String)]]
+  def describeK(algebra: Type): (String, Type => Tree) =
+    "describe" -> { case PolyType(List(f), MethodType(List(af), _)) =>
+      val Af = singleType(NoPrefix, af)
+      val members = overridableMembersOf(Af)
+      val types = delegateAbstractTypes(Af, members, Af)
+      val Tuple2 = symbolOf[Tuple2[Any, Any]]
+      val methods = delegateMethods(Af, members, af) {
+        case method if method.occursInReturn(f) =>
```
joroKr21: We can only implement this if `method.returnType.typeSymbol == f` (same restriction as `productK`). Otherwise the signature won't match.

joroKr21:
Thanks a lot @Fristi and welcome to `cats-tagless`!

I made a PR to your branch that implements the appropriate type for the algebra.

What we still need to decide:
  * Do we want a variant for kind `F[_]` (e.g. `Describe[F[_]]`)? I guess this is probably not that useful.
  * What should we call this? I agree with @kailuowang `DescribeK` is not a great name. But I'm not sure about `Instrument` either. Naming is hard. 
  * I think it would be nice to have a boolean flag `showArguments` of the macro that resolves `cats.Show` for each argument and then returns `methodName(...arguments.map(_.show))`. But that we can defer to another PR.

Otherwise next steps would be:
  * Work on adding an `@autoDescribe` macro annotation
  * Add some tests

Would you be interested to work on this further?

Fristi:
I added
- A simple test case, what tests case could you think of more?
- The className as well in the tuple. I think this is very useful, as algebras can have duplicate method names and therefore to discriminate it's handy to have the algebra name.

joroKr21:
The class name is constant, so you can just use `classOf[Foo]` or directly the string in whatever you do later with the method names. I don't think we should add it, because it only complicates the type signature and generates more code.

About the name, let's use `Instrument` (without K). I couldn't come up with a better one.

About the test, it would be good to add a method with symbolic name.

The other question is if you want to work on an `@autoInstrument` macro annotation (see e.g. `@autoSemigroupalK`. So far we have macro annotations for each method in `Derive`. If not we can leave it for another PR.

Fristi:
I renamed to `Instrument` and added a case class `Instrumentation` which contains the value, algebraName and methodName for now. I also added the `@autoInstrument` annotation which seems to work, but when you add `@finalAlg` it crashes `failed with java.lang.NoClassDefFoundError: cats/tagless/tests/Lookup$`

macros/src/main/scala/cats/tagless/DeriveMacros.scala
```diff
@@ -490,6 +490,30 @@ class DeriveMacros(val c: blackbox.Context) {
...
+      val F = f.asType.toTypeConstructor
+      val Af = singleType(NoPrefix, af)
+      val members = overridableMembersOf(Af)
+      val types = delegateAbstractTypes(Af, members, Af)
+      val algebraName = algebra.typeSymbol.name.decodedName.toString
+
+      val methods = delegateMethods(Af, members, af) {
+        case method if method.returnType.typeSymbol == f =>
+          val body = q"cats.tagless.Instrumentation(${method.body}, $algebraName, ${method.name.decodedName.toString})"
+          val returnType = appliedType(Instrumentation, F :: method.returnType.typeArgs.head :: Nil)
```
joroKr21: Just use `method.returnType.typeArgs`. It's shorter and no need to create a new list.

macros/src/main/scala/cats/tagless/autoInstrument.scala
```diff
@@ -0,0 +1,44 @@
...
+/** Auto generates an instance of `cats.Invariant`. */
+@compileTimeOnly("Cannot expand @autoInstrument")
+class autoInstrument extends StaticAnnotation {
+  def macroTransform(annottees: Any*): Any = macro autoInstrumentMacros.instrumentInst
+}
+
+private[tagless] class autoInstrumentMacros(override val c: whitebox.Context) extends MacroUtils  {
+  import c.universe._
+
+  private def generateInvariantFor(algebraName: String)(algebraType: Tree, typeParams: Seq[TypeDef]) =
```
joroKr21: Please rename this method to `generateInstrumentFor`

tests/src/test/scala/cats/tagless/tests/TestAlgebras.scala
```diff
@@ -113,6 +113,11 @@ object KVStore {
   }
 }
 
+@autoInstrument
+trait Lookup[F[_]] {
```
joroKr21: Let's move this to `object InstrumentTests`

tests/src/test/scala/cats/tagless/tests/InstrumentTests.scala
```diff
@@ -0,0 +1,56 @@
...
+package cats.tagless.tests
+
+import cats.Id
+import cats.tagless.{Derive, Instrument}
+
+class InstrumentTests extends CatsTaglessTestSuite {
+
+
+
+
```
joroKr21: Nitpick: why so much whitespace?

tests/src/test/scala/cats/tagless/tests/TestAlgebras.scala
```diff
@@ -113,6 +113,11 @@ object KVStore {
...
+  def get(id: String): F[Option[Long]]
```
joroKr21: Would you mind adding a symbolic method here and then test the name?
Maybe something like this:
```scala
def ->(id: String): F[Option[String]]
```

Fristi: This fails now on `Referring to non-existent method cats.tagless.tests.InstrumentTests$$anon$4.$$minus$greater(java.lang.String)cats.tagless.Instrumentation`

Fristi:
Fixed all your comments I think. Thanks for the review and help on getting the code to work! Really happy with the fast progress 🎉 

> It worked fine on CI, are you using Intellij?

Yes, any problems? :)

tests/src/test/scala/cats/tagless/tests/TestAlgebras.scala
joroKr21: Again it works on CI. It's Intellij being funky with the tests, no problem otherwise.
In cases like this I just run `test` in `sbt`.

tests/src/test/scala/cats/tagless/tests/InstrumentTests.scala
```diff
@@ -0,0 +1,61 @@
...
+package cats.tagless.tests
+
+import cats.Id
+import cats.tagless.{Derive, Instrument, autoInstrument}
+import InstrumentTests._
+
+class InstrumentTests extends CatsTaglessTestSuite {
+
+  test("Describe should put method and class name in result") {
```
kailuowang: nit pick: the term `Describe` and `desc` should better be renamed to be consistent with the type classname?

joroKr21:
Oh I almost forgot. Do we want to add to the `syntax`?

core/src/main/scala/cats/tagless/Instrument.scala
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless
+
+import simulacrum.typeclass
+
+final case class Instrumentation[F[_], A](
```
kailuowang: Actually, can we name this `Instrumented` since it contains the computed value? By the way I am not married to the term instrument either, I am open to alternatives.

joroKr21:
No rush from my side. One idea I was thinking about next was optionally capturing the arguments.

kailuowang:
Capturing arguments sounds very interesting. Can they be store in special types or just List[Any]?

joroKr21:
Originally I thought about `List[List[String]]` if the parameters have `Show` instances, but `List[List[Any]]` could also work.

Fristi:
Fixed the comments I think :)

Would you mind adding the arguments at a later point? Would love to see `Instrument` this is a release soon if possible as it will me save a ton of code.

My two cents on the arguments. I would prefer arguments to store in a `Map[String, String]` or so where the argument it self has a `Show` instance available. The tricky part is, where do you look for those instances, and what if none around? Do we ignore and warn ? A few of me algebras use refinement types in the argument list. This requires a import which has `Contravariant` instance for these refinement types,

joroKr21:
> Would you mind adding the arguments at a later point? Would love to see Instrument this is a release soon if possible as it will me save a ton of code.

That's also fine by me.

I think I also want to add #51 to the next release but that won't take very long.

kailuowang:
I just one last request, would you move this feature into a new package `cats.tagless.diagnosis` (open to better name)? and add a comment to the type class and macro: `This feature is experimental, API is likely to change`? 
Thanks very much!

dispalt:
I would like to say that I for one am very excited for this.

Fristi:
Fixed the comments

PR #97, opened 2019-12-13
(Inv/Cov/Con)-FunctorK and SemigroupalK instances for core cats classes

cosmin33:
Added instances for:

- **ContravariantK**: Foldable, UnorderedFoldable
- **InvariantK**: Monad, Traverse, FlatMap, Applicative, UnorderedTraverse, UnorderedFoldable, ContravariantMonoidal, Apply, Alternative, ContravariantSemigroupal, MonoidK, InvariantMonoidal, Distributive, SemigroupK, InvariantSemigroupal, Functor, Contravariant, Semigroupal, Invariant
- **FunctorK**: FunctionK[A, *[_]]
- **SemigroupalK**: Monad, Traverse, FlatMap, Applicative, UnorderedTraverse, UnorderedFoldable, ContravariantMonoidal, Apply, Alternative, MonoidK, InvariantMonoidal, Distributive, InvariantSemigroupal, SemigroupK, Functor, Contravariant, Semigroupal, Invariant


joroKr21:
Wow, @cosmin33 thank you very much for the contribution!

This will take some time to review.

But I have a more general question to other maintainers, esp. @kailuowang 
Given typelevel/cats#3043 do we want to move these instances into implicit scope?
We haven't promised binary compatibility yet so we can do it cleanly.

LukaJCB:
@joroKr21 I think that's a great idea!

kailuowang:
I also think it's a good idea to have them in implicit scope.

joroKr21:
@cosmin33 sorry for the inconvenience, would it be too much trouble to move these instances to the companion objects of the typeclasses? You can also do that for the existing instances if you feel like it.

Also I think the implementation traits should be `private[tagless]`. WDYT?
Also I think it would be better to make them abstract classes (or even concrete classes where possible).

joroKr21:
Any ideas how to test the new instances? I think we should be able to test at least the `FunctionK` instance.

cosmin33:
@joroKr21 Sure, I'll move them.
Also, on second thought about the InvariantK instances, I don't think many of them are lawful, they would be only if the functions A => B and B => A constitutes an isomorphism - having the law: 

> to.andThen(from) = identity

 And sadly we don't have an iso in cats yet so we don't have where to put them. So maybe I'll just remove them for now.

joroKr21:
Right, I didn't look closely at the instances but you are right. `Invariant` has no isomorphism requirement.

Optics libraries such as `monocle` usually define isomorphism. Maybe https://github.com/oleg-py/meow-mtl would be a good place for isomorphism-based instances. Or another micro library. I'm not sure.

cosmin33:
ok, closing for repairs...

cosmin33:
Moved all implicits in implicit scope

cosmin33:
One little thing: All those "Any" instances may be changed to use some automatic (macro) helper, as they are very ugly.
Did you explore something like this ? [https://github.com/estatico/scala-cached](url)

joroKr21:
Hi @cosmin33, thanks for the patience. Now that the Christmas holidays are over and slowly getting back to working mode I hope to get to reviewing this soon. In the mean time there are some conflicts.

cosmin33:
@joroKr21, no hurry, thanks for your time; conflicts resolved.

joroKr21:
I agree with moving the instances into implicit scope 👍

joroKr21:
I did some tweaks for consistency. We still need to add tests in another PR

PR #103, opened 2019-12-18
Followup #95 Instrument

joroKr21:
Some things that were missed during review.


joroKr21:
@Fristi @kailuowang WDYT, I think we should get this in before releasing the next version?

Fristi:
LGTM

Issue #29, opened 2019-03-22
Support derivation for case class


Daenyth:
Example:
```scala
case class ServerDependencies[F[_]](
  foo: FooService[F],
  bar: BarService[F],
  other: NotTaglessThing
)
```
I'd expect it to generate something like:
```scala
def mapK[G[_]](f: F ~> G): ServerDependencies[G] =
  ServerDependencies(foo.mapK(f), bar.mapK(f), other)
```

joroKr21:
That's a good point. I've thought about it. It's also trivial to convert a trait with methods to a case class with functions. But I guess it's way more popular to use traits for final tagless in Scala. I'm not sure about any performance implications. 

For case classes we could use shapeless. The problem would be how to integrate that with the traits derivation. Also note that shapeless doesn't support implicit parameters (e.g. `Monad[F]` - it's almost impossible currently). 

Also your example touches upon another issue which is not yet solved - supporting

joroKr21:
Ah, I just realized that shapeless doesn't have `Generic` for this kind of case classes. It has `Generic` for kind `*` and `Generic1` for kind `* -> *` but not for `(* -> *) -> *` which is what we have here. So that would be very difficult!

Krever:
Hey folks, I would need it for higher kinded data. Do you know if there are any blockers to implement this in Scala 3? If it's possible and not too hard I could give it a shot next week.

joroKr21:
> Hey folks, I would need it for higher kinded data. Do you know if there are any blockers to implement this in Scala 3? If it's possible and not too hard I could give it a shot next week.

It should be doable with Shapeless 3. I see two complications:
 * Sealed traits - should we treat them as coproducts via Shapeless or as interfaces via macro?
 * How do we dispatch to either the macro or Shapeless implementation from the same method?

Krever:
Why do you think shapeless is necessary here? I thought we could just use macros.

Also could we (initially at least) not bother ourselves with fully fledged ADTs hierarchies and support only direct case classes/traits?

joroKr21:
> Why do you think shapeless is necessary here? I thought we could just use macros.

It's just much easier and type-safe. I think it's fine to start with case classes only.

joroKr21:
There is even an example in the tests: 
```scala
trait FunctorK[H[_[_]]]:
  def mapK[A[_], B[_]](af: H[A])(f: A ~> B): H[B]

object FunctorK:
[...]

PR #113, opened 2020-01-19
Add TraverseK

kubukoz:
For now, just a general idea & a single law and IdK instance. Ideally, all the laws from Traverse could be borrowed and made into the higher-kinded thing.

Let me know what you think!

### DIFF
--- a/core/src/main/scala/cats/tagless/TraverseK.scala
+++ b/core/src/main/scala/cats/tagless/TraverseK.scala
@@ -1,3 +1,5 @@
+package cats.tagless
+
 /*
  * Copyright 2019 cats-tagless maintainers
  *
@@ -14,15 +16,20 @@
  * limitations under the License.
  */
 
-package cats.tagless
+import simulacrum.typeclass
+import cats.~>
+import cats.Apply
+import cats.arrow.FunctionK
 
-package object syntax {
-  object all extends AllSyntax
-  object functorK extends FunctorKSyntax
-  object contravariantK extends ContravariantKSyntax
-  object invariantK extends InvariantKSyntax
-  object semigroupalK extends SemigroupalKSyntax
-  object applyK extends ApplyKSyntax
-  object instrument extends InstrumentSyntax
-  object traverseK extends TraverseKSyntax
+@typeclass
+trait TraverseK[Alg[_[_]]] extends FunctorK[Alg] {
+
+  def traverseK[F[_], G[_]: Apply, H[_]](alg: Alg[F])(fk: F ~> λ[a => G[H[a]]]): G[Alg[H]]
+
+  def sequenceK[F[_]: Apply, G[_]](alg: Alg[λ[a => F[G[a]]]]): F[Alg[G]] =
+    traverseK[λ[a => F[G[a]]], F, G](alg)(FunctionK.id[λ[a => F[G[a]]]])
+
+  override def mapK[F[_], G[_]](af: Alg[F])(fk: F ~> G): Alg[G] = traverseK[F, cats.Id, G](af)(fk)
+
+  def sequenceKId[F[_]: Apply](alg: Alg[F]): F[Alg[cats.Id]] = sequenceK[F, cats.Id](alg)
 }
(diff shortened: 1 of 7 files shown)


joroKr21:
Maybe I'm missing something but actually I don't see a way to implement this for interfaces.

kubukoz:
I suppose we'd need to call all methods indeed (so it becomes impossible if any of a trait's methods have any parameters)... but it's quite useful for case classes parameterized with a type constructor:

```scala
final case class User[F[_]](id: F[Long], name: F[String])
```

Given an instance, which is relatively straightforward:

```scala
implicit val traverseK: TraverseK[User] = new TraverseK[User] {
  override def traverseK[F[_], G[_]: Apply, H[_]](alg: User[F])(fk: F ~> λ[a => G[H[a]]]): G[User[H]] =
    (fk(alg.id), fk(alg.name)).mapN(User[H])
}
```

we can do things like going from `User[F]` to `F[User[Id]]` (or, more generally, from `User[F compose G]` to `F[User[G]]` - that's `sequenceK`).

So regardless of the fact that we can't do this for interfaces (sadly - I didn't think it through before), I think the algebra itself belongs in a library like cats-tagless and could at least have the derivations for data types.

cc @Fristi

kubukoz:
I also have an OptionT instance now...

```scala
implicit def optionTTraverseK[A]: TraverseK[OptionT[*[_], A]] = new TraverseK[OptionT[*[_], A]] {
      def traverseK[F[_], G[_]: Apply, H[_]](alg: OptionT[F, A])(fk: F ~> λ[a => G[H[a]]]): G[OptionT[H, A]] = fk(alg.value).map(OptionT(_))
    }
```

joroKr21:
We have an open issue for case class support: #29

> I also have an OptionT instance now...
Does that mean we can have an instance for `EitherT` as well? How about `EitherK`, `Tuple2K`?

joroKr21:
This would make sense to revive for Scala 3. But I'm not sure what's the correct signature 🤔

kubukoz:
The one I had in my experiments:

```scala
trait TraverseK[Alg[_[_]]] extends FunctorK[Alg] {

  def traverseK[F[_], G[_]: Apply, H[_]](alg: Alg[F])(fk: F ~> λ[a => G[H[a]]]): G[Alg[H]]

  def sequenceK[F[_]: Apply, G[_]](alg: Alg[λ[a => F[G[a]]]]): F[Alg[G]] =
    traverseK[λ[a => F[G[a]]], F, G](alg)(FunctionK.id[λ[a => F[G[a]]]])

  override def mapK[F[_], G[_]](af: Alg[F])(fk: F ~> G): Alg[G] = traverseK[F, cats.Id, G](af)(fk)

  /** Like [[sequenceK]], but with the second effect hardcoded to [[cats.Id]] for better inference.
    */
  def sequenceKId[F[_]: Apply](alg: Alg[F]): F[Alg[cats.Id]] = sequenceK[F, cats.Id](alg)
}
```

joroKr21:
I find it strange that it has 3 type parameters and that we don't use `ApplyK`.
This also looks pretty strange: `F ~> λ[a => G[H[a]]]`

joroKr21:
Although it does look like `Traversable` from `rank2classes` in Haskell:
https://hackage.haskell.org/package/rank2classes-1.5.3.1/docs/Rank2.html#t:Traversable

joroKr21:
@kubukoz idk about you but I sometimes find myself needing something like this:
```scala
trait FunctorK[T[_[_]]] {
  def mapK[F[_]: Functor, G[_]](t: T[F])(f: F ~> G): T[G]
}
```

So that I can map over stuff like `Expr[F[_]](left: F[Expr[F]], op: InfixOp, right: F[Expr[F]])`.
I wonder how a `TraverseK` would look in this case - I guess:

```scala
trait TraverseK[Alg[_[_]]] extends FunctorK[Alg] {
  def traverseK[F[_]: Traverse, G[_]: Apply, H[_]](alg: Alg[F])(fk: F ~> λ[a => G[H[a]]]): G[Alg[H]]
}
```

PR #121, opened 2020-02-19
NihilistK: create something ex nihilo, from an EmptyK

diesalbla:
DONOTMERGE For some use cases, we may want to create an instance for an Algebra, using an `alleycats.EmptyK`. Essentially, a effectful operation with no values of the inner return type.

#### Example

Given the following algebra: 
```Scala
trait Dwarves[F[_]] { 
  def fili(x: Int): F[Unit]
  def kili(y: Char): F[Boolean] 
}
```

I would like to generate a class that implements that algebra, using in every operation the `empty` operation of the given `EmptyK`:

```Scala
class EmptyDwarves[G[_]](ek: EmptyK[G]) extends Dwarves[G] {
  def fili(x: Int): G[Unit] = ek.empty[Unit]
  def kili(y: Char): G[Boolean] = ek.empty[Boolean]
}
```

The primary use case I am looking for is for testing, to do something similar to what mocks do: I would like to generate an instance of an algebra that, by default, would generate an error in every operation. I would then be able to override any method I wish to implement in each test...

```Scala
import cats.effect.IO

val mockDwarves: EmptyDwarves[IO] = new EmptyDwarves {
  override def fili(x: Int): IO[Unit] = IO.unit
}
```

### DIFF
--- a/core/src/main/scala/cats/tagless/NihilistK.scala
+++ b/core/src/main/scala/cats/tagless/NihilistK.scala
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 cats-tagless maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless
+
+import simulacrum.typeclass
+import alleycats.EmptyK
+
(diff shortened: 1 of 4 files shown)


joroKr21:
Hi that's very interesting. What is the use case for it? Testing?
One fundamental limitation is that we would not be able to override methods selectively.

diesalbla:
@joroKr21 Sorry, left this in mid state. I added a section in the description above.

joroKr21:
> I am looking to see if we can generate classes as well.

This is not possible with blackbox macros. In the current design of cats-tagless we use blackbox macros and a think layer of macro annotations on top of that which delegates to the blackbox macros. You would have to do something special if you want to generate classes.

joroKr21:
I think it would be more interesting to define `type Const[A] = { type λ[B] = A }` and then:
`def const[Alg[_[_]], A](value: A): Alg[Const[A]#λ]`
This together with `Instrument` and `Aspect` (see #141) would give you some mocking capabilities (e.g. by using `Gen`).

diesalbla:
@joroKr21 That may be a better approach, yes. Thanks for taking care of it.

PR #127, opened 2020-03-30
Update Scala versions in Travis config

joroKr21:


joroKr21:
Why is Travis not updating the status here? 🤔

PR #129, opened 2020-03-31
Fix build by updating macro paradise and kind-projector

joroKr21:
Supersedes #128


joroKr21:
The build is green but it looks like Travis is having problem reporting the status on PRs this week 🤔

PR #134, opened 2020-04-15
Fix byname parameters when they have to be transformed

### LINKED ISSUE #133: Trouble deriving instances for ApplicativeError/MonadError
```scala
def invariantKApplicativeError[E]: InvariantK[ApplicativeError[*[_], E]] =
  Derive.invariantK[ApplicativeError[*[_], E]]
```

The above results in:

```scala
could not find implicit value of type cats.tagless.InvariantK[[F[_]]cats.ApplicativeError.CatchOnlyPartiallyApplied[T,F,E]]
```

which doesn't seem too helpful... 

An attempt to derive MonadError:

```scala
def invariantKMonadError[E]: InvariantK[MonadError[*[_], E]] =
  Derive.invariantK[MonadError[*[_], E]]
```

yields 

```scala
could not find implicit value of type cats.tagless.InvariantK[[F[_]]E => F[B]]
```

which is slightly more helpful. Implementing the instance helps, but not a lot:

```scala
implicit def functorKFunctionEffect[I, O]: FunctorK[Lambda[F[_] => I => F[O]]] = new FunctorK[Lambda[F[_] => I => F[O]]] {
      def mapK[F[_], G[_]](af: I => F[O])(fk: F ~> G): I => G[O] = i => fk(af(i))
    }
```

```scala
could not find implicit value of type cats.tagless.InvariantK[[F[_]]=> F[Boolean]]
```

and I implemented that too, but it's not being picked up:

```scala
could not find implicit value of type cats.tagless.InvariantK[[F[_]]=> F[Boolean]]
```

joroKr21:
Also add method name to error message when required implicit is missing.

Fixes #133

### DIFF
--- a/macros/src/main/scala/cats/tagless/DeriveMacros.scala
+++ b/macros/src/main/scala/cats/tagless/DeriveMacros.scala
@@ -86,6 +86,12 @@ class DeriveMacros(val c: blackbox.Context) {
 
     /** The definition of this method as a Scala tree. */
     def definition: Tree = q"override def $name[..$typeParams](...$paramLists): $returnType = $body"
+
+    /** Summon an implicit instance of `A`'s type constructor applied to `typeArgs` if one exists in scope. */
+    def summon[A: TypeTag](typeArgs: Type*): Tree = {
+      val tpe = appliedType(typeOf[A].typeConstructor, typeArgs: _*)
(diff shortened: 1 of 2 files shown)


joroKr21:
Merging this without review because I need it for #141 - can address comments later if any 😄

PR #140, opened 2020-05-02
Update Catalyst and Cats, drop Scala 2.11

joroKr21:
This updates to Scala.js 1.0.

Closes #139 
Closes #138


kailuowang:
I am guessing that you need to remove 2.11 from .travis.yml as well?

joroKr21:
Ah can't default to 2.13 - I forgot about tut.

PR #141, opened 2020-05-03
Add Aspect that can do a form of AOP in Tagless Final with type classes 

### LINKED ISSUE #126: Provide method parameter names and values in Instrument
It would be nice to have access to names and values of the parameters for the given call.

Why would this be useful?

- generating keys for caching
- adding selected parameters as tags to a distributed tracing span
- logging selected parameters in an error handler added to the underlying algebra

This will likely be problematic because of the types, but maybe there's something we can do about this.

The basic idea: for an algebra like

```scala
trait Foo[F[_]] {
  def bar(a: String, b: Int): F[Int]
}
```

I imagine we could have a type like

```scala
trait Parameter[Constraints[_]] {
  type T
  def constraints: Constraints[T]
  def name: String
  def value: T //by name?
}
```

included in `Instrumentation`:

```scala
final case class Instrumentation[F[_], Constraints[_], A](
  value: F[A],
  algebraName: String,
  methodName: String,
  parameters: List[List[Parameter[Constraints]]],
)
```

and somehow define the constraints that we require each parameter to have, at the point of instrumenting:

```scala
val foo: Foo[IO] = ???

//Show[String] and Show[Int] would be required here
val instrumented: Foo[Instrumentation[IO, Show, *]] = foo.instrument[Show]

val result = instrumented.bar("baz", 42)

val stringified = result.parameters.head.map { param =>
  param.name + ": " + param.constraints.show(param.value)
}.mkString(", ") //a: "baz", b: 42
```

I think kinds aren't a problem in this particular case, as parameters will need to be values, so `*`-kinded things (exactly what `Constraints[_]` expects).

For multiple implicits (which I don't think would be a common case, as Show should be enough for most usages), one could define a custom case class with derivations from/to it based on other implicits:

```scala
final case class Constraints[A](show: Show[A], monoid: Monoid[A])
implicit def constraintsToShow[A](implicit constraints: Constraints[A]): Show[A] = ...
```

or something like `EList` (implicit evidence list), as seen in @non's talk: https://youtu.be/O78hnJuzQwA?t=1652

```scala
object foo {
  import cats.implicits._
  type Constraints[A] = Show[A] :/: Monoid[A] :/: ENil

  val constraints: Constraints[String] = ... //some implicit derivation?
}

object bar {
  import foo.constraints

  Show[String] //derives from Constraints[String]
}
```

joroKr21:
The typeclass should have an instance for all (non-implicit) parameters
and also for the return type of every method.

It would be more flexible to allow two typeclasses for arguments and
return type but in practice this will be always `Show`, `Encoder` and
similar. On the other hand it would make the API too ugly.

Also add a `Trivial` typeclass that is always available.

Opening this for feedback. I'm still not sure about the API and naming.
CC @kubukoz 

Fixes #126


core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
```diff
@@ -24,12 +24,52 @@ final case class Instrumentation[F[_], A](
   methodName: String
 )
 
+object Instrumentation {
+  final case class With[F[_], G[_], A](
+    instrumentation: Instrumentation[F, A],
+    arguments: List[List[Argument[G]]],
```
kubukoz, line 30: I'd rename G to something like Instances.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
```diff
@@ -24,12 +24,52 @@ final case class Instrumentation[F[_], A](
...
+  )
+
+  object With {
+    def instance[F[_], G[_], A](
+      instrumentation: Instrumentation[F, A],
+      arguments: List[List[Argument[G]]]
+    )(implicit G: G[A]): With[F, G, A] = With(instrumentation, arguments, G)
+  }
+
+  sealed trait Argument[F[_]] extends Serializable {
```
kubukoz, line 41: Same point about naming

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
```diff
@@ -24,12 +24,52 @@ final case class Instrumentation[F[_], A](
...
+    type A
+    def name: String
+    def value: A
```
kubukoz, line 44: Does this work with by-names?

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
```diff
@@ -24,12 +24,52 @@ final case class Instrumentation[F[_], A](
...
   * @tparam Alg The algebra type you want to instrument.
   */
 @typeclass
-trait Instrument[Alg[_[_]]] {
+trait Instrument[Alg[_[_]]] extends Serializable {
   def instrument[F[_]](af: Alg[F]): Alg[Instrumentation[F, *]]
 }
+
+object Instrument {
+  trait With[Alg[_[_]], G[_]] extends Serializable {
```
kubukoz, line 72: Mixed feelings about having a single type for instances on return types and parameters.

I can think of cases in which you might want `Monoid` of the return type and `Show` for the parameters. Even stacking two `With` on top of each other won't give you that power, I'm afraid.

I think we could make the API support using two different types, but at the same type expose a higher-level variant that does both at once, too.

kubukoz, line 72: Is this `@typeclass` material?

docs/src/main/tut/index.md
```diff
@@ -17,7 +17,7 @@ Cats-tagless is a small library built to facilitate composing tagless final enco
 
 ## Installation
 
-Cats-tagless is available on scala 2.11, 2.12, 2.13 and scalajs. 
+Cats-tagless is available on scala 2.12, 2.13 and scalajs.
```
kubukoz, line 20: the README doesn't mention SJS, should we add that there too?

core/src/main/scala/cats/tagless/syntax/InstrumentSyntax.scala
```diff
@@ -16,6 +16,19 @@
...
-trait InstrumentSyntax extends Instrument.ToInstrumentOps
+trait InstrumentSyntax extends Instrument.ToInstrumentOps {
+  import InstrumentSyntax.InstrumentConstrainedOps
+
+  implicit def toInstrumentConstrainedOps[Alg[_[_]], F[_]](af: Alg[F]): InstrumentConstrainedOps[Alg, F] =
+    new InstrumentConstrainedOps(af)
+}
+
+object InstrumentSyntax {
+  class InstrumentConstrainedOps[Alg[_[_]], F[_]](private val af: Alg[F]) extends AnyVal {
```
kubukoz, line 29: Some leftovers from past renaming, I think (Constrained?)

kubukoz:
Oh yeah, and I miss `F[_], A[_], R[_], T[_]`.

macros/src/main/scala/cats/tagless/Derive.scala
```diff
@@ -66,4 +66,5 @@ object Derive {
     * @return An instrumented algebra.
     */
   def instrument[Alg[_[_]]]: Instrument[Alg] = macro DeriveMacros.instrument[Alg]
+  def instrumentWith[Alg[_[_]], G[_]]: Instrument.With[Alg, G] = macro DeriveMacros.instrumentWith[Alg, G]
```
kubukoz, line 69: Is an annotation possible too? `@autoInstrumentWith[Show]`

kubukoz:
> The typeclass should have an instance for all (non-implicit) parameters

I'd consider having instances for implicits too, but maybe only ones that don't look like a typeclass... on the other hand, maybe it's useless, so take the suggestion with a grain of salt.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: I'm not sure about that, it doesn't sound great. Instance is usually a value not a type.
And if we decide to go for two typeclasses then they will need different names.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: Hmm not right now, I guess will have to make it a lazy val.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: Yes, it's strictly less powerful.
It would be greatly appreciated if you have an example use case.
The biggest challenge will be naming things 😄

joroKr21: `@typeclass` works only with one type parameter

docs/src/main/tut/index.md
joroKr21: I will add in in #140 👍

core/src/main/scala/cats/tagless/syntax/InstrumentSyntax.scala
joroKr21: Yeah 😄 (I didn't like the sound of it)

macros/src/main/scala/cats/tagless/Derive.scala
joroKr21: Some research is necessary. I don't know if we can get access to `Show`

kubukoz:
Here's the diff from my POC with shapeless 😂  (`59 additions and 265 deletions`)

https://github.com/kubukoz/talks/compare/cats-tagless-instrumentwith-demo

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
kubukoz: well, it'll be the type of instances :)

I agree about needing two names. But then `G` and `H` will be even less meaningful.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
kubukoz: oh, I guess it can still be a def. Especially since by-name parameters aren't memoized.

I guess I forgot `def` isn't `val` 😅

joroKr21:
> I'd consider having instances for implicits too, but maybe only ones that don't look like a typeclass... on the other hand, maybe it's useless, so take the suggestion with a grain of salt.

But how would we know? This would be too hand-wavy and so far we managed to keep cats-tagless quite principled 😄

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
kubukoz: I linked this example https://github.com/kubukoz/talks/compare/cats-tagless-instrumentwith-demo in which I'm using `Instrument.With` with `Show` and `Semigroup` in two distinct places, but I'm using standard types and have all the instances. But it isn't a stretch to think you'd have an algebra in your business code that takes parameters that don't have (or even cannot have) Semigroup/Monoid/whatnot when the return type does, or the other way around...

kubukoz: fair enough

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: It matters how we construct an `Argument` though

kubukoz:
Yeah, I imagine that'd require creating escape hatches (like annotations for parameters to ignore, or forcing users to implement the instance themselves if they want to customize it). Let's ignore implicits for now and we can get back to them if anyone asks with a real case.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: The last thing I tried before that was `With[F[_], Arg[_], Ret[_], A]`

kubukoz: That sounds good.

Have you thought about the case where someone might want more than one TC for arguments / return types? E.g. both a Semigroup and Show for args, but Encoder for return types.

joroKr21: >Have you thought about the case where someone might want more than one TC for arguments / return types? E.g. both a Semigroup and Show for args, but Encoder for return types.

They could just create a combined typeclass.
It could also be generalized by having something like `HList1`.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: That looks like more than instrumentation though.
Maybe there is some orthogonal concept we are missing.
I need to sleep on it, going to bed now 😄

joroKr21:
@kubukoz I just pushed a version that permits different typeclasses for arguments and return value, the API is quite more complicated now.

kubukoz:
I think it's okay :) do you think `WithBoth` could be `Full`? The "both" part doesn't really say what the two things are.

core/src/main/scala/cats/tagless/diagnosis/Instrument.scala
joroKr21: There is some orthogonality but parameter names mess it up.
I'd say let's keep it specific to instrumentation for now.
The orthogonality would be constrained typeclasses.

joroKr21:
`Full` doesn't sound great either 🤔

Fristi:
I checked out this stuff yesterday, looks great!

For naming maybe use the terminology around functions? So

- `Codomain` - For function return types
- `Domain` - For function argument types
- `Function` - For both 

And since it's not really only instrumenting and it's recalls me of aspect weaving it might be better to name it `Weaving` using an `Aspect` (type class) which yields a `Weave` (now `Instrumentation`)

Aside from that, beside naming any blockers? Would love to see this released :)

joroKr21:
@Fristi indeed no blockers besides naming! I will give your ideas a try today or tomorrow. Also need to add some tests.

joroKr21:
Cool use case - given `Aspect[Foo, Decoder, Encoder]` we can define `Foo[Endpoint]` - think HTTP endpoints / routes.

Fristi:
Interesting idea! Using these techniques algebras can be woven into remote variants

joroKr21:
Yes there would be still be some manual boilerplate left. I guess it's better suited on the client side than the server side.

core/src/main/scala/cats/tagless/aop/Aspect.scala
```diff
@@ -0,0 +1,103 @@
...
+/** Type class supporting Aspect-Oriented Programming (AOP) for a tagless final algebra.
+  * The behaviour of this aspect is defined by two type classes (or constraints) -
+  * one for the domain (arguments) and one for the codomain (result) of each method in the algebra.
+  *
+  * Note: This feature is experimental, API is likely to change.
+  *
+  * @see `Aspect.Domain`, `Aspect.Codomain` and `Aspect.Function` for convenient type aliases.
+  * @tparam Alg The algebra to which this aspect applies.
+  * @tparam Dom Type class / constraint that should exist for all parameter types of each algebra methods (domain).
+  * @tparam Cod Type class / constraint that should exist for the return type of each algebra methods (codomain).
```
kubukoz, line 31: Sounds fishy 😆

core/src/main/scala/cats/tagless/aop/Aspect.scala
```diff
@@ -0,0 +1,103 @@
...
+object Aspect {
+  type Domain[Alg[_[_]], F[_]] = Aspect[Alg, F, Trivial]
+  type Codomain[Alg[_[_]], F[_]] = Aspect[Alg, Trivial, F]
+  type Function[Alg[_[_]], F[_]] = Aspect[Alg, F, F]
+
+  /** An [[Aspect.Weave]] represents a reified cross-cutting concern for a single method of an algebra.
+    * It can be applied to all method arguments, result, or both. Its behaviour is driven by type classes.
+    *
+    * @see `Weave.Domain`, `Weave.Codomain` and `Weave.Function` for convenient type aliases.
+    * @param domain `Aspect.Advice` for all arguments of the method. Target in `Eval` to capture by-name arguments.
```
kubukoz, line 48: Does it include implicits? Might be worth mentioning.

kubukoz: Also, if so, I would type it differently, so that it's possible to tell whether the argument list is an implicit one. There can only be one (in Scala 2 at least) so something like

```scala
final case class ArgumentLists(base: List[List[Advice[Eval, Dom]], implicits: Option[List[Advice[Eval, Dom]]])
```

comes to mind.

kubukoz: It could actually be a different field in Weave.

joroKr21:
The last commit overhauls the naming to use AOP terminology.
I think it makes more sense now, although I don't feel good about defining `type Function`.

I also did some changes to `Instrument` and made a arranged `FunctorK`, `Instrument` and `Aspect` into a hierarchy.
I think all use cases require a `FunctorK` anyway and the conditions to derive them are a subset.
We can also now add laws that `Instrument` and `Aspect` don't change semantics (`fa.instrument.mapK(_.value) <-> fa`),
but I would prefer to do that in another PR because this one is getting already quite big.

core/src/main/scala/cats/tagless/aop/Aspect.scala
joroKr21: Do you think `Codom` is a better name? 😄

kubukoz: oh my god

core/src/main/scala/cats/tagless/aop/Aspect.scala
joroKr21: I still can't think of when you would want to include implicit arguments.
There would almost never be an instance of any type class for them.
Also this would break the use case for generic methods (see the test for `Show aspect`).

kubukoz: Sure - it makes sense. I would just mention that they aren't included :)

Fristi:
Looks great :)

joroKr21:
Ahh realised something annoying - having multiple implicit aspects in scope will make `FunctorK` and `Instrument` ambiguous.

Fristi:
I also see you moved `algebraName` to `Instrument`. I guess thats more effecient in terms of performance. For me it's not convenient.

See the accompanied example

```scala

val woven = new PostgresRepos[Aspect.Weave.Codomain[RIO[OpenTracing, *], TraceData, *]] {
   // when call `Derive.aspect.algebraName` you can access it's name, but it doesn't make sense here yet
   def depots = Derive.aspect.weave(repos.depots)
  // more repos here
}


val traceTransformation = new ((Aspect.Weave.Codomain[RIO[OpenTracing, *], TraceData, *]) ~> RIO[OpenTracing, *]) {
            def apply[A](fa: Codomain[RIO[OpenTracing, *], TraceData, A]): RIO[OpenTracing, A] = {
              fa.codomain.target.flatMap { res =>
                fa.codomain.instance.extract(res) match {
                  case Some(value) =>
                    // here I also would like to access algebra name, to add it to the tag
                    OpenTracing.tag("result", value) *> ZIO.succeed(res)
                  case None => ZIO.succeed(res)
                }
              }.span(fa.codomain.name)
            }
          }

          woven.mapK(traceTransformation)

```

PR #150, opened 2020-07-27
update sbt-catalysts and Scala versions

larsrh:


larsrh:
Another instance of scala-js/scala-js#4122, caused by https://github.com/typelevel/sbt-catalysts/blob/v0.39/src/main/scala/sbtcatalysts/CatalystsPlugin.scala#L472

PR #153, opened 2020-07-29
Dependency injection via ReaderT

joroKr21:
I guess mostly ZIO users would be happy about this but it cost almost nothing to implement.


joroKr21:
I'm going to merge this, but I'd be happy to address any feedback or even revert later.

PR #155, opened 2020-08-05
Bring back algebraName at Instrumentation and add it to Aspect

Fristi:
It's quite convenient to have the `algebraName` at the site when transforming the algebra:

```scala
val woven = new PostgresRepos[Aspect.Weave.Codomain[RIO[OpenTracing, *], TraceData, *]] {
   // when call `Derive.aspect.algebraName` you can access it's name, but it doesn't make sense here yet
   def depots = Derive.aspect.weave(repos.depots)
  // more repos here
}

val traceTransformation = new ((Aspect.Weave.Codomain[RIO[OpenTracing, *], TraceData, *]) ~> RIO[OpenTracing, *]) {
  def apply[A](fa: Codomain[RIO[OpenTracing, *], TraceData, A]): RIO[OpenTracing, A] = {
    fa.codomain.target.flatMap { res =>
      fa.codomain.instance.extract(res) match {
        case Some(value) =>
          // here I also would like to access algebra name, to add it to the tag
          OpenTracing.tag("result", value) *> ZIO.succeed(res)
        case None => ZIO.succeed(res)
      }
    }.span(fa.codomain.name)
  }
}

woven.mapK(traceTransformation)
```

### DIFF
--- a/core/src/main/scala/cats/tagless/aop/Aspect.scala
+++ b/core/src/main/scala/cats/tagless/aop/Aspect.scala
@@ -53,13 +53,14 @@ object Aspect {
     * @tparam A Return type of the method.
     */
   final case class Weave[F[_], Dom[_], Cod[_], A](
+    algebraName: String,
     domain: List[List[Advice[Eval, Dom]]],
     codomain: Advice.Aux[F, Cod, A]
   ) {
 
     /** Convert this [[Weave]] to an [[Instrumentation]], throwing away information about the `domain`. */
     def instrumentation: Instrumentation[F, A] =
-      Instrumentation(codomain.target, codomain.algebraName, codomain.methodName)
+      Instrumentation(codomain.target, algebraName, codomain.name)
   }
 
   object Weave {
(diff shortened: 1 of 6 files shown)


joroKr21:
I think it should be part of `Weave`, not of `Advice`.

Fristi:
Yea good point! Should the method name also be in there ?

joroKr21:
`Advice.name` the method name in `codomain` and parameter name in `domain`.

Fristi:
Makes sense! Fixed it

core/src/main/scala/cats/tagless/aop/Aspect.scala
```diff
@@ -86,10 +87,10 @@ object Aspect {
   object Advice {
     type Aux[F[_], G[_], T] = Advice[F, G] { type A = T }
 
-    def apply[F[_], G[_], T](adviceName: String, adviceTarget: F[T])(implicit G: G[T]): Aux[F, G, T] =
+    def apply[F[_], G[_], T](method: String, adviceTarget: F[T])(implicit G: G[T]): Aux[F, G, T] =
```
joroKr21, line 90: Let's revert this change because for `domain` it's the argument's name
```suggestion
    def apply[F[_], G[_], T](adviceName: String, adviceTarget: F[T])(implicit G: G[T]): Aux[F, G, T] =
```

core/src/main/scala/cats/tagless/aop/Aspect.scala
```diff
@@ -86,10 +87,10 @@ object Aspect {
...
       new Advice[F, G] {
         type A = T
-        val name = adviceName
+        val name = method
```
joroKr21, line 93:
```suggestion
        val name = adviceName
```

PR #162, opened 2020-08-24
Add Two New Error Related Methods To FunctorK.

isomarcte:
This commit adds two new methods on the `FunctorK` typeclass, `transformError` and `flatTapOnError`.

For any `FunctorK[A]` for which the current `F` on `A[F]` is an instance of `MonadError[F, E]` (for some `E`) we can provide a method which allows for transformation on all `E` values for all methods in the given `A[F]`. This uses the identity `FunctionK` for invocations which do not yield an error and the `handleErrorWith` methods from `MonadError` for the error paths.

The primary motivating use case for these methods is creating a succinct way to define logging in the case of an error for all methods of an F-Algebra. For example with these changes it becomes possible to write something like this,

```scala
trait Logger[F[_]] {
    def error(message: String): F[Unit]
    def warn(message: String): F[Unit]
    def info(message: String): F[Unit]
}

@autoFunctorK
trait FAlgebra[F[_]] {
    def lookupUser(id: Id): F[User]
    def lookupOrder(orderId: OrderId): F[Order]
    def createOrder(order: Order): F[OrderId]
}

object FAlgebra {
    def simpleInstance[F[_]]: FAlgebra[F] = new FAlgebra[F]{
        override def lookupUser(id: Id): F[User] = ???
        override def lookupOrder(orderId: OrderId): F[Order] = ???
        override def createOrder(order: Order): F[OrderId] = ???
    }

    def withErrorLogging[F[_]: Sync](baseInstance: FAlgebra[F], logger: Logger[F]): FAlgebra[F] =
        baseInstance.flatTapOnError{
            case t: Throwable =>
                logger.error(t.getMessage)
        }
}

```

### DIFF
--- a/core/src/main/scala/cats/tagless/FunctorK.scala
+++ b/core/src/main/scala/cats/tagless/FunctorK.scala
@@ -16,18 +16,39 @@
 
 package cats.tagless
 
+import cats._
+import cats.arrow._
+import cats.implicits._
 import simulacrum.typeclass
-import cats.~>
 
 /**
- * Sort of a higher kinded Functor, but, well, it's complcated. 
+ * Sort of a higher kinded Functor, but, well, it's complcated.
  * See Daniel Spiewak's comment here
  * https://github.com/typelevel/cats/issues/2697#issuecomment-453883055
- * Also explains why this isn't in `cats-core`. 
-**/ 
+ * Also explains why this isn't in `cats-core`.
+**/
 @typeclass
 trait FunctorK[A[_[_]]] extends InvariantK[A] {
   def mapK[F[_], G[_]](af: A[F])(fk: F ~> G): A[G]
 
   override def imapK[F[_], G[_]](af: A[F])(fk: F ~> G)(gK: G ~> F): A[G] = mapK(af)(fk)
+
+  /** Perform a transformation on all errors. */
+  final def transformError[F[_], E](af: A[F])(f: E => F[E])(implicit F: MonadError[F, E]): A[F] =
+    mapK(af)(
+      new FunctionK[F, F]{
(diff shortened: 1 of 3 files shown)


kailuowang:
This seems useful, but to be honest, I am not quite sure if these methods belong to the type class `FunctorK`. My personal opinion is that it makes more sense to add the functionality of generating such a `FunctionK` to `MonadError`.  It might take some time to play with different api designs to test the ergonomics.

isomarcte:
@kailuowang yeah I wasn't entirely sure if this is the correct place for them either.

I'm not exactly sure how'd we add these to `MonadError` since we need the `FunctorK` which isn't in `cats-core`, unless I'm misunderstanding you.

kailuowang:
@isomarcte the way I see it, the method we add to MonadError doesn't not involve mapK, rather it simply returns a `FunctionK`, then it should be easy to combine that with `mapK`.

isomarcte:
@kailuowang Oh I see what you mean. You mean that we generate the `FunctionK` for these and then just use `mapK` manually to invoke them.

I'd be okay with that. Would you like me to open such a PR on `cats-core` and close this one?

joroKr21:
I agree with @kailuowang it's too specific to add to the `FunctorK` type class. There are probably many functions in Cats that could have a `*K` variant and I'm not sure what is the best approach. If a PR is accepted in core, they should be added there. If not I'm not sure. There could be an object `FunctionKs` or they could live in user code. At $work we had such a `Transformations` object with functions, some of which were quite project specific. E.g. you might want to change the error type, use lenses, etc.

isomarcte:
Closing this because it doesn't really fit into cats-core as it doesn't really have a use case that makes sense beyond FunctorK.

PR #182, opened 2020-12-04
Call out need for -Ymacro-annotations

keirlawson:
Its mentioned in the README.md but we should also call it out here as well.


joroKr21 (approved):
That makes sense, thank you 🙏

PR #196, opened 2021-01-24
Scala 2.12.13

joroKr21:


kailuowang (approved):

joroKr21:
Looks like scoverage is not compatible

PR #203, opened 2021-02-16
Replace simulacrum with the expanded source code

joroKr21:
First step towards Scala 3 support.
Code generated by passing `-Dsimulacrum.trace` to sbt and formatted.
simulacrum-scalafix doesn't work for us due to typelevel/simulacrum-scalafix#100
I tried to solve that problem but it requires more effort than I was willing to sink.


kailuowang (approved):
LGTM

PR #227, opened 2021-04-06
Fix root project and publish snapshots

joroKr21:


build.sbt
```diff
@@ -63,11 +65,11 @@ val macroSettings = List(
   })
 )
 
-lazy val catsTagless = project
+lazy val `cats-tagless` = project
+  .in(file("."))
```
joroKr21, line 69: 😭 why can't this stuff just work

PR #314, opened 2022-01-21
Migrate to sbt-typelevel

armanbilge:


armanbilge:
```
[error] core: Failed binary compatibility check against org.typelevel:cats-tagless-core_2.12:0.14.0 (e:info.apiURL=https://typelevel.org/cats-tagless/api/, e:info.versionScheme=early-semver)! Found 2 potential problems
[error]  * class cats.tagless.implicits.package does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.tagless.implicits.package")
[error]  * object cats.tagless.implicits.package does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.tagless.implicits.package$")
```

armanbilge:
Seems this change in https://github.com/typelevel/cats-tagless/pull/273/ was binary incompatible:
https://github.com/typelevel/cats-tagless/pull/273/files#diff-20bc798be15502fa5e6a3f636d7a49dfd2e93d17f8f08b0913fa909a2eb23beb

Is it intended for your next breaking version?

armanbilge:
Just fixed some merge conflicts. Any thoughts on this? :)

joroKr21:
I guess the binary breaking change is not intentional but on the other hand we're pre-1.0 so not much compatibility guarantees. If there's an easy way to fix though perhaps it's worth it.

armanbilge:
Yep, easy to fix.

core/src/main/scala/cats/tagless/implicits.scala
```diff
@@ -16,4 +16,9 @@
 
 package cats.tagless
 
-object implicits extends syntax.AllSyntax
+import scala.annotation.nowarn
+
+package implicits {
+  @nowarn("cat=deprecation") // required for bincompat
+  object `package` extends syntax.AllSyntax
+}
```
armanbilge, line 24: Ok, this was not so easy actually. But, miraculously it works 😁

armanbilge:
@joroKr21 good-to-go 👍

core/src/main/scala/cats/tagless/implicits.scala
joroKr21: Ah ok I remember now why - it's deprecated and because of `-Xsource:3`

armanbilge: Yes, and annoying `@nowarn` doesn't work for `package object` for some reason.

joroKr21: Wait, does that mean `package object implicits` doesn't work with `nowarn`?

armanbilge: Yes, that's why I rewrote it in this strange way.

joroKr21: I'm surprised that this works though - it doesn't break source code?

armanbilge: Yes, I was too. But AFAICT your mdocs use the`import` and it they compile fine. Maybe scalac special-cases this? The pattern was important enough that there's a test case for it:
https://github.com/adriaanm/scala/blob/3e18e9fe50dbbfba62e413ff6c404121b1a5f29b/test/files/neg/deprecate_package_object_extends.scala

joroKr21: Wow, mind blown - thanks for digging this out :exploding_head:

joroKr21 (approved):
Thanks again!

PR #323, opened 2022-02-06
Add license to sbt

joroKr21:
See https://github.com/sbt/sbt/releases/tag/v1.6.2
This is just adding the existing license to the build.

@armanbilge hmm any idea why it's complaining?
I noticed that it worked fine for you in typelevel/kittens#429

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -191,6 +191,7 @@ lazy val commonNativeSettings = List(
 )
 
 ThisBuild / homepage := Some(url(homePage))
+ThisBuild / licenses := Seq(License.Apache2)
 ThisBuild / developers := List(
   Developer(
     "Georgi Krastev",


armanbilge:
sbt-typelevel already sets Apache 2.0 as the default  (not sbt-typelevel-ci-release though).

https://github.com/typelevel/sbt-typelevel/blob/8a88b6ea1a91b95e67a4fed68029e8c8e7cee1e1/core/src/main/scala/org/typelevel/sbt/TypelevelPlugin.scala#L57

Although I see now I may need to fix the link.

joroKr21:
I just noticed from the sbt release notes that there is a predefined `License.Apache2` and it works.
So should I close this PR as unnecessary then?

armanbilge:
Up to you :) it's harmless but effectively redundant (modulo the license url which I'll fix).

joroKr21:
Taken care by `sbt-typelevel`

Issue #125, opened 2020-03-29
Having multiple annotations on a trait without a companion throws at runtime


kubukoz:
```scala
import cats.tagless.implicits._
import cats.tagless.finalAlg
import cats.tagless.autoFunctorK
import cats.arrow.FunctionK

//or autoFunctorK + autoInstrument, or finalAlg + autoInstrument, etc.
@autoFunctorK
@finalAlg
trait Foo[F[_]] {
  def bar(a: Int): F[Int]
}

//object Foo {}

object Main extends App {
  val foo: Foo[cats.Id] = _ => 42

  println(foo.mapK(FunctionK.id))
}
```

This will result in `java.lang.NoClassDefFoundError: Foo$`. Uncommenting the companion object fixes things.

joroKr21:
Ha, interesting. I guess the two annotations are competing on adding the companion object 🤔

dispalt:
yup ran into this a bunch before, real annoying could never figure out a workaround.

joroKr21:
But the tests are full of such algebras with two annotations and no companion objects:
https://github.com/typelevel/cats-tagless/blob/master/tests/src/test/scala/cats/tagless/tests/autoFunctorKTests.scala

Could it be something related to incremental compilation?
Can you try to clean and compile again?

dispalt:
I forgot to mention but it works when its within another object, another reason why it was so annoying to figure it out.  If you move one of those traits with two annotations to the top level of the file, it won't work, IIRC.

danslapman:
Good old bug: https://github.com/scalameta/scalameta/issues/1046, this must be fixed in scalac (since paradise was merged in 2.13)

joroKr21:
Would this fix the problem? scala/scala#9555

PR #328, opened 2022-02-15
Documentation: Warning and resolution for multiple annotations

msinton:
I'm using cats-tagless for the first time and hit this issue:
https://github.com/typelevel/cats-tagless/issues/125

I think this must be common enough to merit living in documentation until resolved.


joroKr21:
Thanks @msinton that's a great idea!

Unfortunately the issue itself comes from the Scala compiler so we can't solve it here, but adding this to the FAQ is a great addition!

joroKr21 (approved):

PR #382, opened 2022-10-06
Add a variant of `readerT` that works for methods that return `Stream[F, A]`

m50d:
This probably isn't suitable for merging as-is, but opening for discussion: we have a use case where we want to derive instances where some of the methods return fs2 `Stream`, and `ReaderT` is particularly difficult. Is this something cats-tagless would be open to supporting? Maybe there's a simpler way to express this?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -110,7 +110,10 @@ lazy val macros = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(
     moduleName := "cats-tagless-macros",
     scalacOptions := scalacOptions.value.filterNot(_.startsWith("-Wunused")).filterNot(_.startsWith("-Ywarn-unused")),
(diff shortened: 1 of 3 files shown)


joroKr21:
It sounds reasonable but there are some practical problems, the most notable of which is having a dependency on fs2.
I wonder if this can be generalised somehow to depend on type classes from Cats but it doesn't look to be the case.

m50d:
It might be possible to express it in terms of `FunctorK`, and then provide an instance of `FunctorK[Stream[?, A]]` in some downstream module (which is probably a good idea anyway)? I'm not sure if there's a cats typeclass that represents the fact that we can go `F ~> Stream[F, ?]` (the equivalent of `MonadTrans#liftM` in scalaz) which is the other piece that that approach would need.

joroKr21:
I don't think `FunctorK` is enough though - you are also using `flatMap` in this PR

PR #401, opened 2023-01-15
Initial Scala 3 support

pomadchin:
Connected https://github.com/typelevel/cats-tagless/issues/170

**TODO (an approximate list)**:

Supported instances derivation body functions:

- **FunctorK**: https://github.com/typelevel/cats-tagless/pull/442
  - [x] `A => F[_]`
  - [ ] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [ ] `Kleisli / WriterT / WriterT / etc` => `A`
  - [x] `AlgWithNonEffectMethod`
  - [ ] `AlgWithTypeMember`
  - [x] `AlgWithTypeBound`
  - [x] `[X] => F[X] => F[_]` generic functions support
    - [x] `AlgWithTParamInMethod` 
  - [ ] `AlgWithAbstractTypeClass`
  - [x] `AlgWithCurryMethod`
  - [ ] Algebras with extra Type Parameters 
    - [ ] `AlgWithExtraTP`
    - [x] `AlgWithExtraTP2` 
  - [x] Varargs
    - [x]  `AlgWithVarArgsParameter`
  - [ ] `BuilderAlgebra`
- **SemigroupalK** https://github.com/typelevel/cats-tagless/pull/513, https://github.com/typelevel/cats-tagless/pull/516
  - [x] `A => F[_]`
  - [ ] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [ ] `Kleisli / WriterT / WriterT / etc` => `A`
  - [x] `algWithGenericType`
  - [x] `algWithCurryMethod`
  - [x] Varargs
    - [x]  `AlgWithVarArgsParameter`
- **ApplyK** https://github.com/typelevel/cats-tagless/pull/516
  - [x] `A => F[_]` functions support 
  - [ ] Varargs
    - [ ]  `AlgWithVarArgsParameter`
- **InvariantK** https://github.com/typelevel/cats-tagless/pull/513
  - [x] `A => F[_]`
  - [ ] `A => U[F]`
  - [ ] `F[_] => F[_]`
  - [ ] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [ ] `Kleisli / WriterT / WriterT / etc` => `A`
  - [ ] `NoEffectMethod`
  - [ ] `ContravariantEff`
  - [ ] `InvariantEff`
  - [ ] `AlgWithTypeMember`
  - [ ] Algebras with extra Type Parameters 
[...]

Issue #170, opened 2020-11-08
Build for Dotty


larsrh:

joroKr21:
https://scalacenter.github.io/scala-3-migration-guide/docs/macros/migration-tutorial.html

smarter:
This library uses macro annotations which don't exist in Dotty, but things like `@autoFunctorK trait Foo` should be rewritable as `trait Foo derives FunctorK` using https://dotty.epfl.ch/docs/reference/contextual/derivation.html. In fact it looks like [`shapeless-3`](https://github.com/milessabin/shapeless/tree/shapeless-3) already has support for that: (there's an example in the readme)

joroKr21:
@smarter the macro annotations are not essential. They just generate an implicit instance in the companion object but the heavy lifting is done by blackbox macros. If we want to use shapeless 3 we would need to somehow generate mirrors for traits.

Fristi:
First of all, great work with this library.

Just doing some evaluation for scala 3. We use it in some projects and we would like to know if we could use it when Scala 3 comes out? Or do you plan to port the most used macros first? `FunctorK` and `Tupled2K` are essential, aspect weavng is nice but can be added later.

All the macros need to be rewritte right? I am not an expert :) Contributed on by looking at the examples of the work of the maintainers. Let me know if I can help a hand.

joroKr21:
> Just doing some evaluation for scala 3. We use it in some projects and we would like to know if we could use it when Scala 3 comes out? Or do you plan to port the most used macros first? FunctorK and Tupled2K are essential, aspect weavng is nice but can be added later.

It's on my todo list to work on this. I can't promise that it will be done by the time Scala 3 final out though. But - if we manage to port one macro, we can port them all. The implementation is very consistent. The only odd ones are `Instrument` and `Aspect`. Note that there are no macro annotations on Scala 3 so they will have to be replaced by support for `derives FunctorK` etc. clauses.

 > All the macros need to be rewrite right? I am not an expert :) Contributed only by looking at the examples of the work of the maintainers. Let me know if I can help a hand.

Embarrassingly, the most demotivating part of the work is setting up the build and CI for cross publishing 🙈 We also need to apply [simulacrum-scalafix](https://github.com/typelevel/simulacrum-scalafix) before we can start because we use `@typeclass`. So here are two self-contained tasks we need to do first:
 1. Apply `simulacrum-scalafix`
[...]

PR #425, opened 2023-05-25
Initial Core Scala 3 cross compilation support

pomadchin:
Connected https://github.com/typelevel/cats-tagless/issues/170

This PR is an attempt to split https://github.com/typelevel/cats-tagless/pull/401 into a separate PR and add the core project cross compilation.

1. No scala 3 macro code, no scala 3 tests
2. Old tests moved into the scala-2 dir


DoSOfRedRiver:
Looking good! 
@joroKr21 can we get this merged and Scala 3 build released? Really need this update 🙏

core/src/main/scala-2/cats/tagless/derived/DerivedApplyK.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless.derived
+
+trait DerivedApplyK
```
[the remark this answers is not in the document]
joroKr21: I guess one issue with that approach is that we can't add it to type classes that are defined in Cats proper. I think that's fine but I just wanted to mention it. For those we can use extension methods to add the `derived` instance.

core/src/main/scala/cats/tagless/FunctionKLift.scala
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package cats.tagless
+
+import cats.arrow.FunctionK
+
+// https://github.com/typelevel/cats/issues/2553#issuecomment-493712879
+// https://github.com/typelevel/cats/blob/v2.9.0/core/src/main/scala-2/src/main/scala/cats/arrow/FunctionKMacros.scala
+object FunctionKLift {
```
joroKr21, line 23: I don't get it - why do we need to copy this here? We are using `liftFunction` already.

core/src/main/scala-2/cats/tagless/derived/DerivedApplyK.scala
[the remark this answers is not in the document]
pomadchin: @joroKr21 wow I actually didn't know that; your comment btw https://github.com/typelevel/cats/pull/3820#issuecomment-826267914 

Hmm, I can remove those traits so we defer the decision until later.

core/src/main/scala/cats/tagless/InvariantK.scala
```diff
@@ -28,63 +29,67 @@ trait InvariantK[Alg[_[_]]] extends Serializable {
   def imapK[F[_], G[_]](af: Alg[F])(fk: F ~> G)(gk: G ~> F): Alg[G]
 }
 
-object InvariantK extends InvariantKInstances01 {
-  implicit def catsTaglessApplyKForEitherK[F[_], A]: ApplyK[EitherK[F, *[_], A]] =
-    eitherKInstance.asInstanceOf[ApplyK[EitherK[F, *[_], A]]]
+object InvariantK extends InvariantKInstances01 with DerivedInvariantK {
+  implicit def catsTaglessApplyKForEitherK[F[_], A]: ApplyK[({ type W[g[_]] = EitherK[F, g, A] })#W] =
+    eitherKInstance.asInstanceOf[ApplyK[({ type W[g[_]] = EitherK[F, g, A] })#W]]
```
joroKr21, line 34: Can we use `-Ykind-projector` on Scala 3?
https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html#rewriting-incompatible-constructs

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
```diff
@@ -0,0 +1,56 @@
...
+package discipline
+
+import org.scalacheck.Arbitrary
+import org.scalacheck.Prop.*
+import cats.{Eq, ~>}
+import cats.data.Tuple2K
+import cats.laws.discipline.*
+import cats.tagless.laws.discipline.SemigroupalKTests.IsomorphismsK
+
+trait ApplyKTests[F[_[_]]] extends FunctorKTests[F] with SemigroupalKTests[F] {
```
joroKr21, line 28: Sorry if it's a stupid question, since we have no diff here - why did you have to split the test files into different versions?

build.sbt
```diff
@@ -184,7 +188,20 @@ lazy val commonSettings = List(
...
+    case Some((3, _)) =>
+      Seq(
+        "-language:adhocExtensions",
+        "-source:future",
+        "-explain"
+      )
+    case Some((2, 12 | 13)) => Seq("-Xsource:3", "-P:kind-projector:underscore-placeholders")
+    case _ => Nil
+  }),
+  // sbt-typelevel sets -source:3.0-migration, we'd like to replace it with -source:future
```
joroKr21, line 202: Sorry, I've bee out for a while. What was the difference between `source:future` and `source:3.0-migration`? If you have a link to read about it I would be grateful 🙏

build.sbt
```diff
@@ -184,7 +188,20 @@ lazy val commonSettings = List(
...
+  scalacOptions := scalacOptions.value.filterNot(_ == "-source:3.0-migration"),
```
joroKr21, line 203: I have since learned about it:
```suggestion
  scalacOptions ~= (_.filterNot(_ == "-source:3.0-migration")),
```

joroKr21 (commented):
This is a massive effort - thank you so much for taking this on! I'm sorry for the delays on my side but I think I finally have a bit of a more stable environment so let's get the ball rolling 👍

core/src/main/scala/cats/tagless/InvariantK.scala
pomadchin: I think I tried that, but let me double check. I think I had a cross compilation issue, sry that is a bit old PR from the past, I forgot some of the context

core/src/main/scala-2/cats/tagless/derived/DerivedApplyK.scala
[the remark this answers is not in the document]
joroKr21: I think it's still fine - let's keep them. Just wanted to point it out.

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
pomadchin: A super dumb solution ): 

in pre 2.12, we could write, all around the function application and `_`:

```scala
val props = List("applyK associativity" -> forAll(laws.applyKAssociativity[A, B, C] _))
```

Scala 3 wants to see it as 

```scala
val props = List("applyK associativity" -> forAll(laws.applyKAssociativity[A, B, C]))
```

pomadchin: If there is a better way to solve it I'm happy to apply it :+1:

core/src/main/scala/cats/tagless/InvariantK.scala
joroKr21: No worries, I can also try it

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
joroKr21: You mean we had to do it? Ahh, there has to be a common subset 😄

core/src/main/scala/cats/tagless/FunctionKLift.scala
joroKr21: Oh shoot, they reached some kind of decision on access, but it was a different decision than Scala 2:
https://github.com/scala/bug/issues/6794

Let's make this private for now and advocate for Cats to make it public.

pomadchin: Scala3 lift:

```scala
def lift[F[_], G[_]](f: [X] => F[X] => G[X]): FunctionK[F, G] =
    new FunctionK[F, G] {
      def apply[A](fa: F[A]): G[A] = f(fa)
    }
```

Scala2 lift:

```scala
def liftFunction[F[_], G[_]](f: F[τ[F, G]] => G[τ[F, G]]): FunctionK[F, G] =
    new FunctionK[F, G] {
      def apply[A](fa: F[A]): G[A] = f.asInstanceOf[F[A] => G[A]](fa)
    }
```

For cross compilation we need
1. We need the same signature (well, at least the invocation should be similar)
2. We need the same name, liftFunction is not present in Scala3

See the question from the older PR: https://github.com/typelevel/cats-tagless/pull/401#discussion_r1070750299

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
pomadchin: yea I could not find a common subset between scala 2.12, 2.13, and 3; 
If we drop 2.12 we can just drop underscores!

core/src/main/scala/cats/tagless/FunctionKLift.scala
joroKr21: Oh wow, I'm contradicting myself now. I've changed my mind I guess 😄 
The issue with polymorphic functions is that they have no type inference.
It's been a long time now and it's still not implemented. It would be good to have a solution in the meantime.

core/src/main/scala/cats/tagless/InvariantK.scala
pomadchin: I bet there was smth with multiple args in a type constructor and cross compilation, I think I could not achieve a nice syntax there, multiple type params or is not supproted, or I could not make it work.

In Scala 3 we could use full type lambdas with `=>>` and all that stuff.

joroKr21: Ohh I read it totally wrong - it says it **doesn't** support higher kinds, not even with `λ`
This is a disappointment, but now I remember 😢

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
joroKr21: The common subset is to use `laws.applyKAssociativity[A, B, C](_, _, _)` - I know it's not satisfactory, but it's better than duplicating the entire file.

joroKr21:
@pomadchin I think we should keep `macros` as a Scala 2 only project

build.sbt
pomadchin: I was trying to find a reference, can't find a good link.

But the long story short, we need it for all of those experimental macros from the larger PR.
It basically means what it says - compiler is aggressive on the future new syntax usage only, new language features are enabled.

Basically you can close the old PR, rm this flag and see how everything is not compiling.

pomadchin:
@joroKr21 also don't hesitate just to push into my branch, I think you should have permissions just to push inside. The privilege of being a project maintainer :D

core/src/main/scala/cats/tagless/FunctionKLift.scala
pomadchin: @joroKr21 I'm all on board to advocate it being available in the scala3 cats! Count me in to throw some extra support in comments.

build.sbt
joroKr21: Ok, got it 👍

pomadchin: 🤦 clo**s**e => clo**n**e

joroKr21:
> also don't hesitate just to push into my branch, I think you should have permissions just to push inside. The privilege of being a project maintainer :D

Alright, thanks 👍 But I don't want to step on your toes either. I will try an idea to get rid of the ugly kind-projector stuff.

laws/src/main/scala-2.13+/cats/tagless/laws/discipline/ApplyKTests.scala
pomadchin: https://github.com/typelevel/cats-tagless/pull/425/commits/afae45851fc20287c0402482d79261aca38820e2 🎉

core/src/main/scala/cats/tagless/FunctionKLift.scala
pomadchin: Made it private for now https://github.com/typelevel/cats-tagless/pull/425/commits/afae45851fc20287c0402482d79261aca38820e2

pomadchin:
🤔 Hm, with making macro a scala2 project only can be a bit tricky, `tests` depend on it.

joroKr21:
> 🤔 Hm, with making macro a scala2 project only can be a bit tricky, tests depend on it. We definitely need to exclude it from publishing.

Ok, that's a good alternative

joroKr21:
I'm falling asleep already - but please let me know what you think of the `Curried` type aliases

pomadchin:
> I'm falling asleep already - but please let me know what you think of the `Curried` type aliases

I think it is good! Definitely less ugly. So sad that we could not keep it as lambdas ): 

I won't be available this weekend (up until Tuesday), no-laptop long weekend; but will be here to help Tuesday morning!

build.sbt
```diff
@@ -44,24 +45,34 @@ ThisBuild / githubWorkflowAddedJobs ++= Seq(
...
   })
 )
 
+val macroPublishSettings = List(
+  publish / skip := (CrossVersion.partialVersion(scalaVersion.value) match {
+    case Some((3, _)) => true
+    case _ => false
+  })
+)
+
```
pomadchin, line 75: This is the publish skip 🤔 There are definitely some SBT wizards that can make it better, but this change is very convenient for now.

joroKr21 (approved):

DoSOfRedRiver:
Is there anything that stop us from releasing Scala 3 build?

joroKr21:
@DoSOfRedRiver this build only the core module, i.e. no automatic derivation. Is that useful to you?
If so I think there's nothing stopping us.

DoSOfRedRiver:
Yep, that is what I was looking for =)
I currently working on an update for another library, which is heavily depends on cats-tagless. No macros support needed at the moment though.

joroKr21:
Ok, I create a release https://github.com/typelevel/cats-tagless/releases/tag/v0.15.0
It should eventually make its way to maven automatically.

pomadchin:
Thanks for figuring it out that fast! 🔥

DoSOfRedRiver:
Thanks a lot! Already making use of it =)

core/src/main/scala/cats/tagless/FunctionKLift.scala
pomadchin: => https://github.com/typelevel/cats/issues/4449

PR #442, opened 2023-06-09
Scala3 FunctorK macros

pomadchin:
This PR adds the Scala3 FunctorK macro prototype.

 Any help with ensuring that all tests pass are welcome.

There is also some extra Scala macro helpers functions added, that are not really used in terms of this PR, so it's up to us -- drop it for now or keep it alive. 🤷 

Supported instances derivation body functions:

- **FunctorK**:
  - [x] `A => F[_]`
  - [x] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [x] `Kleisli / WriterT / WriterT / etc` => `A`
  - [x] `AlgWithNonEffectMethod`
  - [ ] `AlgWithTypeMember`
  - [x] `AlgWithTypeBound`
  - [x] `[X] => F[X] => F[_]` generic functions support
    - [x] `AlgWithTParamInMethod` 
  - [ ] `AlgWithAbstractTypeClass`
  - [x] `AlgWithCurryMethod`
  - [x] Algebras with extra Type Parameters 
    - [x] `AlgWithExtraTP`
    - [x] `AlgWithExtraTP2` 
  - [x] Varargs
    - [x]  `AlgWithVarArgsParameter`
  - [x] `BuilderAlgebra`

Connected https://github.com/typelevel/cats-tagless/pull/401, #170


tests/src/test/scala-3/cats/tagless/tests/TestAlgebras.scala
```diff
@@ -0,0 +1,156 @@
...
+import cats.laws.discipline.arbitrary.*
+import cats.laws.discipline.eq.*
+import cats.syntax.all.*
+import cats.{Eq, Eval, Monoid, ~>}
+import org.scalacheck.{Arbitrary, Cogen}
+
+import scala.util.Try
+
+// TODO: @finalAlg @autoProductNK @autoInstrument
+trait SafeAlg[F[_]] derives FunctorK: //, SemigroupalK:
```
pomadchin, line 31: I know it's ugly, but I want to propose to keep those things commented out as much as possible so we'll do ~the same test coverage 😞

tests/src/test/scala-3/cats/tagless/tests/autoFunctorKTests.scala
```diff
@@ -0,0 +1,348 @@
...
+import cats.arrow.FunctionK
+import cats.data.Cokleisli
+import cats.free.Free
+import cats.laws.discipline.SerializableTests
+import cats.tagless.laws.discipline.FunctorKTests
+
+import scala.annotation.nowarn
+import scala.util.Try
+
+class autoFunctorKTests extends CatsTaglessTestSuite:
```
pomadchin, line 49: This is the Suite we want to ensure is 100% covered to match the old macro behavior.
Or at least 90%, with some documented limitations 🦐

tests/src/test/scala-3/cats/tagless/tests/simple/FunctorKSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+import cats.tagless.*
+import cats.tagless.syntax.all.*
+import cats.tagless.macros.*
+
+import cats.Id
+import cats.arrow.FunctionK
+
+import scala.compiletime.testing.*
+
+class FunctorKSpec extends munit.FunSuite with Fixtures:
```
pomadchin, line 28: Simple is very very useful for the macro development and macro experiments.

tests/src/test/scala-3/cats/tagless/tests/TestAlgebras.scala
joroKr21: I agree 👍

tests/src/test/scala-3/cats/tagless/tests/autoFunctorKTests.scala
joroKr21: Yeah it may be difficult to get 100% coverage. Admittedly some of the functionality is quite niche.

pomadchin:
`AlgWithTypeMember` is an interesting one, it can't prove that the passed from the outside given instance matches to the expected by the internal function one.

`AlgWithExtraTP` feels like _a feature_, mb worth rising a compiler issue / question.

`BuilderAlgebra` not supported now, macro code should be adjusted.

tests/src/test/scala-3/cats/tagless/tests/autoFunctorKTests.scala
```diff
@@ -0,0 +1,345 @@
...
+
+  trait AlgWithNonEffectMethod[F[_]] derives FunctorK:
+    def a(i: Int): F[Int]
+    def b(i: Int): Int
+
+  // TODO: @finalAlg
+  // trait AlgWithTypeMember[F[_]] derives FunctorK {
+  //   type T
+  //   def a(i: Int): F[T]
+  // }
```
pomadchin, line 244: The root cause of it is that we generate now smth like:

```scala
    new FunctorK[AlgWithTypeMember] {
      def mapK[F[_], G[_]](af: AlgWithTypeMember[F])(fk: F ~> G): AlgWithTypeMember[G] = {
        new AlgWithTypeMember[G] {
          def a(i: Int): G[T] = fk(af.a(i))
        }
      }
    }
```

What we should:

```scala
    new FunctorK[AlgWithTypeMember] {
      def mapK[F[_], G[_]](af: AlgWithTypeMember[F])(fk: F ~> G): AlgWithTypeMember[G] = {
        new AlgWithTypeMember[G] {
          type T = af.T
          def a(i: Int): G[T] = fk(af.a(i))
        }
      }
    }
```

How to do `type T = af.T` via macro -- that's the question 🤔

joroKr21: Oh wow, they haven't exposed the method to create a `TypeDef` symbol - what's the point of exposing incomplete APIs like this?

pomadchin: @joroKr21 yea, this morning (I forgot to reply u) I played a bit with it; 

It's ez to deconstruct

```scala
TypeTree.of[Alg[G]].tpe.typeSymbol.typeMembers.filter(_.isTypeDef).map(_.tree).collect {
  case td @ TypeDef(, tree) => td
}
```

But I was not able to "construct" one that points to `Alg[F].T`

joroKr21: And why is even `TypeDef.apply` there, how can it be used? 😂

pomadchin: If I hardcode it to `type T = Int`:

```scala
TypeDef(T,TypeTree[TypeBounds(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),object scala),class Int),TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),object scala),class Int))])
```

pomadchin: 🤔 mb we can use copy somehow?

```scala
trait TypeDefModule { this: TypeDef.type =>
  def apply(symbol: Symbol): TypeDef
  def copy(original: Tree)(name: String, rhs: Tree): TypeDef
  def unapply(tdef: TypeDef): (String, Tree)
}
```

pomadchin: `apply` here could be for the empty tree o:

joroKr21: But you need to pass the declarations to `Symbol.newClass`

pomadchin: Not really! 
`Symbol.newClass` makes a new class symbol only; so if we build a correct `TypeDef` somehow we can put it into the class body already (that's the code in this repo); check the `ClassDef`:

```scala
val cls = Symbol.newClass(...)

val tmembers: List[Statement] = 
  TypeTree.of[Alg[G]].tpe.typeSymbol.typeMembers.filter(_.isTypeDef).map(t => (t, t.tree)).collect {
    case (tpe, td @ TypeDef(_, typedTree)) => 
      // we cool and did construct correct TypeDefs somehow
  }
  
val body: List[Statement] = ??? // construct methods

// fill in the body class
val clsDef = ClassDef(cls, parents, body = (body ::: tmembers))

// constructor
val newCls = Typed(Apply(Select(New(TypeIdent(cls)), cls.primaryConstructor), Nil), TypeTree.of[Alg[G]])

// mk an expression
val expr = Block(List(clsDef), newCls).asExpr
expr.asExprOf[Alg[G]]
```

pomadchin: ^ in fact we can do it even now, but I was not able to properly ref type member of `Alg[G]` to the `Alg[F].T` 😭

joroKr21: Here is how it is:
```scala
val cls = Symbol.newClass(Symbol.spliceOwner, className, parents = parents.map(_.tpe), decls, selfType = None)
    val body =
      cls.declaredMethods.map(method => (method, method.tree)).collect { case (method, DefDef(_, _, typedTree, _)) =>
        DefDef(
          method,
          argss =>
            typedTree.tpe.simplified.asMatchable match
              case at @ AppliedType(o, inner) =>
                val apply = methodApply(eaf)(method, argss)
                Some(Select.overloaded(efk.asTerm, "apply", inner, List(apply)))
              case e =>
                val apply = methodApply(eaf)(method, argss)
                Some(apply)
        )
      }
```

You create the symbols in `newClass` (given itself as the owner) and then take them from `declaredMethods` to use them to mage the actual method bodies.

pomadchin: well yea, but we can still do `cls.typeMembers`

pomadchin: but yea ): like there is a missing constructor or smth, no way / I dont see it now to construct TypeDefs

joroKr21: Well you're doing the entire `asSeenFrom` stuff manually so assume you didn't have another choice 😄 
So the same would apply to type members.

pomadchin: ): yea

pomadchin: but mb not; it's not a type param of a trait... 
but hm lets say we don't care about it right now... We still should be able to manually create a `TypeDef` with a ref to `alga.T`

pomadchin: 😞

joroKr21: The `TypeDef` needs to have a symbol and every symbol has an owner. The owner is the class symbol. But a class symbol needs to have its children entered into a scope. Hence they are created in this function which gives you the class symbol as a parameter. It's kinda cyclic.

core/src/main/scala-3/cats/tagless/macros/Utils.scala
```diff
@@ -23,60 +23,54 @@ import quoted.*
...
-
-  def methodApply[Alg[_[_]]: Type, F[_]: Type](e: Expr[Alg[F]])(using Quotes)(
-      method: quotes.reflect.Symbol,
-      argss: List[List[quotes.reflect.Tree]]
-  ): quotes.reflect.Term =
-    import quotes.reflect.*
+private object Utils:
+  // Unfortunately there is no flag for default parameters.
+  val defaultRegex = ".*\\$default\\$\\d+".r
+  def make(using q: Quotes): Utils[q.type] = new Utils
```
pomadchin, line 29: this is a life changer 🙏

tests/src/test/scala-3/cats/tagless/tests/autoFunctorKTests.scala
joroKr21: @pomadchin I just had a thought before going to bed - maybe we can export the types.

joroKr21: Yes, bug again we're lacking the capability to construct them 😄

pomadchin: fantastic :D

pomadchin: I'll try to find some time to make an isolated demo to ask folks in the dotty discussions, mb we'll get some guidance there!

pomadchin:
ouf fmt is unhappy + experimental is more strict now; I hope I return back to this PR soon.

core/src/main/scala-3/cats/tagless/macros/macroFunctorK.scala
```diff
@@ -0,0 +1,63 @@
...
+            .appliedTo(fk.asTerm)
+      },
+      body = {
+        case (tpe, body) if tpe.contains(g) =>
+          Select
+            .unique(tpe.summonLambda[FunctorK](g), "mapK")
+            .appliedToTypes(List(F, G))
+            .appliedTo(body)
+            .appliedTo(fk.asTerm)
+      }
```
pomadchin, line 62: that's wow

joroKr21: I was just gonna mention you that I did some work today and at least `FunctorK` looks viable now.
The only significant part missing is the type members but we're blocked on that.
Now we should be able to implement other type classes more easily.

pomadchin: Do you know what's the status of type members? Is there a related github issue?

joroKr21: https://github.com/scala/scala3/issues/19448

tests/src/test/scala/cats/tagless/tests/OptimizerTests.scala
```diff
@@ -20,9 +20,11 @@ import cats.*
 import cats.data.*
 import cats.tagless.ApplyK
 import cats.tagless.optimize.*
-import cats.tagless.tests.Interpreters.StateInfo
+import scala.annotation.experimental
 
+@experimental
 class OptimizerTests extends CatsTaglessTestSuite {
+  import cats.tagless.tests.Interpreters.StateInfo
```
joroKr21, line 27: Gah I wish we could use Scala 3.4.0 with the `-experimental` flag but something's breaking Scala Native although it says it should support 3.4.0

core/src/main/scala-3/cats/tagless/macros/macroFunctorK.scala
joroKr21: Do you think you will be able to port the rest of your work from #401 to use these helper methods?
If you think so I would merge this to book the progress.

pomadchin: I think I will be able to! Do you want me to make PRs one by one? Meaning one type class at a time.

pomadchin: ship this one to main!! 🔥

joroKr21: > I think I will be able to! Do you want me to make PRs one by one? Meaning one type class at a time.

Whatever works for you

PR #443, opened 2023-06-09
Add FunctorK instance for FS2 Stream

m50d:
One of the pieces for a better attempt at #382 ; this could also live in a separate project, but I think it's small enough for here?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -90,7 +90,8 @@ lazy val fs2 = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(rootSettings)
   .settings(
     moduleName := "cats-tagless-fs2",
-    libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version
+    libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version,
+    tlMimaPreviousVersions := Set()
   )
 
 lazy val lawsJVM = laws.jvm
--- a/build.sbt
+++ b/build.sbt
@@ -91,7 +91,7 @@ lazy val fs2 = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(
     moduleName := "cats-tagless-fs2",
     libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version,
-    tlMimaPreviousVersions := Set()
+    tlVersionIntroduced := Map("2.12" -> "0.15.1", "2.13" -> "0.15.1", "3" -> "0.15.1")
   )
 
 lazy val lawsJVM = laws.jvm
--- a/build.sbt
+++ b/build.sbt
@@ -134,7 +134,7 @@ lazy val macros = crossProject(JVMPlatform, JSPlatform, NativePlatform)
     publish / skip := scalaBinaryVersion.value.startsWith("3"),
     tlMimaPreviousVersions ++= when(scalaBinaryVersion.value.startsWith("2"))("0.14.0").toSet,
     mimaPreviousArtifacts := when(scalaBinaryVersion.value.startsWith("2"))(mimaPreviousArtifacts.value.toSeq*).toSet,
-    tlVersionIntroduced := Map ("3" -> "0.15.0")
+    tlVersionIntroduced := Map("3" -> "0.15.0")
   )
 
 lazy val testsJVM = tests.jvm
--- a/build.sbt
+++ b/build.sbt
@@ -45,6 +45,7 @@ val catsVersion = "2.9.0"
 val circeVersion = "0.14.5"
 val disciplineVersion = "1.5.1"
 val disciplineMunitVersion = "2.0.0-M3"
+val fs2Version = "3.7.0"
 val kindProjectorVersion = "0.13.2"
 val paradiseVersion = "2.1.1"
 val scalaCheckVersion = "1.17.0"
@@ -61,7 +62,7 @@ val macroSettings = List(
   )
 )
 
-lazy val root = tlCrossRootProject.aggregate(core, laws, tests, macros, examples)
+lazy val root = tlCrossRootProject.aggregate(core, fs2, laws, tests, macros, examples)
 
 lazy val coreJVM = core.jvm
 lazy val coreJS = core.js
@@ -75,6 +76,21 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(
     moduleName := "cats-tagless-core",
     libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion
+  )
+
+lazy val fs2JVM = fs2.jvm
+lazy val fs2JS = fs2.js
+lazy val fs2Native = fs2.native
+lazy val fs2 = crossProject(JVMPlatform, JSPlatform, NativePlatform)
+  .crossType(CrossType.Pure)
+  .dependsOn(core)
+  .enablePlugins(AutomateHeaderPlugin)
+  .jsSettings(commonJsSettings)
+  .nativeSettings(commonNativeSettings)
+  .settings(rootSettings)
+  .settings(
+    moduleName := "cats-tagless-fs2",
+    libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version
   )
 
 lazy val lawsJVM = laws.jvm
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,9 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(rootSettings)
   .settings(
     moduleName := "cats-tagless-core",
-    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion
+    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion,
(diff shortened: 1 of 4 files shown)


joroKr21:
Three questions:
 1. I wonder if it's possible to keep it in core (and thus in implicit scope) by declaring the dependency on `fs2` as `Optional`. This is the biggest question we need to answer. We would just have to make sure it doesn't cause issues for users that don't depend on `fs2`.
 2. What other instances can we provide? Perhaps `SemigroupalK` or even `ApplyK`?
 3. Tests - for `FunctorK` this is not blocking because the implementation is trivial. For more complex implementations it might be necessary.

joroKr21:
On the other hand if you plan to work on delivering the functionality from #382 of `ReaderT` including streams, then I think it's better to keep it as the separate module 👍

m50d:
> I wonder if it's possible to keep it in core (and thus in implicit scope) by declaring the dependency on fs2 as Optional. This is the biggest question we need to answer.

> On the other hand if you plan to work on delivering the functionality from https://github.com/typelevel/cats-tagless/pull/382 of ReaderT including streams (which involves macros), then I think it's better to keep it as the separate module +1

I'm not an SBT expert by any means. I think Optional would be ok, but yeah my ultimate goal is to achieve what #382 does.

> What other instances can we provide? Perhaps SemigroupalK or even ApplyK?

I looked at that but couldn't find a way to fit the semantics for `SemigroupalK`, since the effects need to correspond 1:1 wheras for a stream we might need to evaluate more effects from one side than the other to pull the next element, if that makes sense (and we can't require `F` and `G` to be monadic and still fit the definition of `SemigroupalK`). Maybe I'm missing something.

> Tests - for FunctorK this is not blocking because the implementation is trivial. For more complex implementations it might be necessary.

Agreed.

build.sbt
```diff
@@ -77,6 +78,22 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
...
+  .crossType(CrossType.Pure)
+  .dependsOn(core)
+  .enablePlugins(AutomateHeaderPlugin)
+  .jsSettings(commonJsSettings)
+  .nativeSettings(commonNativeSettings)
+  .settings(rootSettings)
+  .settings(
+    moduleName := "cats-tagless-fs2",
+    libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version,
+    tlMimaPreviousVersions := Set()
```
bpholt, line 94: I think a better way to do this would be to set `tlVersionIntroduced`; e.g. 
```scala
tlVersionIntroduced := Map("2.12" -> "0.15.1", "2.13" -> "0.15.1", "3" -> "0.15.1")
```

That way subsequent versions will start checking for compatibility using MiMa.

m50d: This seems to have not quite worked somehow? Any idea what I'm missing?

bpholt: Alright I can reproduce the failure if I run `project rootJVM; ++ 2.12.18; mimaReportBinaryIssues` on the sbt console. 👀

bpholt: I think the issue is caused by the `tlVersionIntroduced` and `tlMimaPreviousVersions` settings in `commonSettings`. I was a little worried something like this would happen when [I added `tlMimaPreviousVersions` to `commonSettings`](https://github.com/typelevel/cats-tagless/pull/430/files), but I left it like that because I didn't want to introduce too big of a change without a good reason. Now that it has caused trouble, I think we can correct it.

If we apply the following diff, I think it will work going forward:
```diff
diff --git a/build.sbt b/build.sbt
index 6ec6409..ffb8ee0 100644
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,9 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(rootSettings)
   .settings(
     moduleName := "cats-tagless-core",
-    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion
+    libraryDependencies += "org.typelevel" %%% "cats-core" % catsVersion,
+    tlVersionIntroduced := Map("3" -> "0.15.0"),
+    tlMimaPreviousVersions ++= when(scalaBinaryVersion.value.startsWith("2"))("0.14.0").toSet
   )
 
 lazy val fs2JVM = fs2.jvm
@@ -109,7 +111,9 @@ lazy val laws = crossProject(JVMPlatform, JSPlatform, NativePlatform)
     libraryDependencies ++= List(
       "org.typelevel" %%% "cats-laws" % catsVersion,
       "org.typelevel" %%% "discipline-core" % disciplineVersion
-    )
+    ),
+    tlVersionIntroduced := Map("3" -> "0.15.0"),
+    tlMimaPreviousVersions ++= when(scalaBinaryVersion.value.startsWith("2"))("0.14.0").toSet
   )
 
 lazy val macrosJVM = macros.jvm
@@ -128,7 +132,9 @@ lazy val macros = crossProject(JVMPlatform, JSPlatform, NativePlatform)
     scalacOptions ~= (_.filterNot(opt => opt.startsWith("-Wunused") || opt.startsWith("-Ywarn-unused"))),
     libraryDependencies += "org.scalacheck" %%% "scalacheck" % scalaCheckVersion % Test,
     publish / skip := scalaBinaryVersion.value.startsWith("3"),
-    mimaPreviousArtifacts := when(scalaBinaryVersion.value.startsWith("2"))(mimaPreviousArtifacts.value.toSeq*).toSet
+    tlMimaPreviousVersions ++= when(scalaBinaryVersion.value.startsWith("2"))("0.14.0").toSet,
+    mimaPreviousArtifacts := when(scalaBinaryVersion.value.startsWith("2"))(mimaPreviousArtifacts.value.toSeq*).toSet,
+    tlVersionIntroduced := Map ("3" -> "0.15.0")
   )
 
 lazy val testsJVM = tests.jvm
@@ -212,8 +218,6 @@ lazy val commonSettings = List(
   startYear := Some(2019),
   apiURL := Some(url("https://typelevel.org/cats-tagless/api/")),
   autoAPIMappings := true,
-  tlVersionIntroduced := Map("3" -> "0.15.0"),
-  tlMimaPreviousVersions ++= when(scalaBinaryVersion.value.startsWith("2"))("0.14.0").toSet,
   // sbt-typelevel sets -source:3.0-migration, we'd like to replace it with -source:future
   scalacOptions ~= (_.filterNot(_ == "-source:3.0-migration")),
   scalacOptions ++= (scalaBinaryVersion.value match {
```

build.sbt
```diff
@@ -74,7 +75,25 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
...
+  .crossType(CrossType.Pure)
+  .dependsOn(core)
+  .enablePlugins(AutomateHeaderPlugin)
+  .jsSettings(commonJsSettings)
+  .nativeSettings(commonNativeSettings)
+  .settings(rootSettings)
+  .settings(
+    moduleName := "cats-tagless-fs2",
+    libraryDependencies += "co.fs2" %%% "fs2-core" % fs2Version,
+    tlVersionIntroduced := Map("2.12" -> "0.15.1", "2.13" -> "0.15.1", "3" -> "0.15.1")
```
joroKr21, line 96: I think it's better to be 0.16.0 - it's not really a bugfix patch release.
```suggestion
    tlVersionIntroduced := Map("2.12" -> "0.16.0", "2.13" -> "0.16.0", "3" -> "0.16.0")
```

bpholt: I disagree—if we were past 1.x, that would be right, but since we're pre-1.x, we're in early semver mode, meaning the second position is effectively a major version bump (no implied binary compatibility), and the third is for both bug fixes and new features. 

Similarly, the `v0.15.0` release should really have been `v0.14.1`. Bumping the version to `v0.15.0` meant sbt started raising errors when evicting `v0.14.0`, even though we know they're binary compatible. I've had a bunch of build failures as a result—Scala Steward makes a PR updating cats-tagless, but a bunch of our apps use cats-tagless along with other libraries that pull cats-tagless in transitively. The Steward PRs that only update cats-tagless fail because of the sbt eviction errors unless everything is updated at once.

It's too late to do anything about `v0.15.0` now, but we can avoid future headaches by using `v0.15.1` as the next version, only going to `v0.16.0` if we need to introduce binary-incompatible changes.

joroKr21: Hmm, fair point - I wasn't aware of any libraries that pull in cats-tagless. Let's keep it `0.15` for now.
We can release 1.0 once we have are done with Scala 3 support 👍

bpholt: Yeah, I ran with [one of your suggestions a while ago](https://github.com/typelevel/cats-tagless/pull/250) and built [Dwolla/async-utils](https://github.com/dwolla/async-utils) for safely lifting tagless-final algebras implemented in Future to cats-effect. I've also created [Dwolla/natchez-tagless](https://github.com/dwolla/natchez-tagless), which contains helpers to more easily use the AOP functionality here with Natchez tracing. (I plan to do something similar for otel4s once it's a little further along.) Our [Dwolla/fs2-aws](https://github.com/Dwolla/fs2-aws) library publishes cats-tagless typeclass instances for some of the algebras it makes available.

joroKr21: That's really cool, thanks for the references!

PR #513, opened 2024-03-26
Scala3 SemigroupalK and InvariantK macros

pomadchin:
- **SemigroupalK**
  - [x] `A => F[_]`
  - [ ] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [ ] `Kleisli / WriterT / WriterT / etc` => `A`
  - [x] `algWithGenericType`
  - [x] `algWithCurryMethod`
  - [x] Varargs
    - [x]  `AlgWithVarArgsParameter`
- **InvariantK**
  - [x] `A => F[_]`
  - [ ] `A => U[F]`
  - [ ] `F[_] => F[_]`
  - [ ] `A => Kleisli / WriterT / WriterT / etc`
    - is there a generic solution?
  - [ ] `Kleisli / WriterT / WriterT / etc` => `A`
  - [ ] `NoEffectMethod`
  - [ ] `ContravariantEff`
  - [ ] `InvariantEff`
  - [ ] `AlgWithTypeMember`
  - [ ] Algebras with extra Type Parameters 
    - [ ] `WithExtraTpeParam`
    - [ ] `AlgWithExtraTP2` 
  - [ ] `AlgWithDef`
  - [ ] `AlgWithTypeParam`
  - [ ] `AlgWithCurryMethod`
  - [ ] Varargs
    - [ ]  `AlgWithVarArgsParameter`
  - [ ] `AlgWithByNameParameter`

Mb I was too hasty and it makes sense to deal with invariant in a separate PR.

### DIFF
--- a/core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2019 cats-tagless maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless.macros
+
+import cats.tagless.*
+import cats.data.Tuple2K
+
+import scala.annotation.experimental
+import scala.quoted.*
+
+@experimental
+object MacroSemigroupalK:
+  inline def derive[Alg[_[_]]] = ${ semigroupalK[Alg] }
+
+  def semigroupalK[Alg[_[_]]: Type](using Quotes): Expr[SemigroupalK[Alg]] = '{
+    new SemigroupalK[Alg]:
+      def productK[F[_], G[_]](af: Alg[F], ag: Alg[G]): Alg[Tuple2K[F, G, *]] =
+        ${ deriveProductK('af, 'ag) }
+  }
+
+  def deriveProductK[Alg[_[_]]: Type, F[_]: Type, G[_]: Type](
+      eaf: Expr[Alg[F]],
+      eag: Expr[Alg[G]]
+  )(using q: Quotes): Expr[Alg[Tuple2K[F, G, *]]] =
+    import quotes.reflect.*
+    given DeriveMacros[q.type] = new DeriveMacros
+
+    val F = TypeRepr.of[F]
+    val G = TypeRepr.of[G]
+    val H = TypeRepr.of[Tuple2K[F, G, *]]
+    val f = F.typeSymbol
+    val h = H.typeSymbol
+
+    List(eaf.asTerm, eag.asTerm).transform[Alg[Tuple2K[F, G, *]]] {
+      case (tpe, argf :: argg :: Nil) if tpe.contains(h) =>
+        Select
+          .unique(tpe.typeArgs.head.appliedTo(tpe.typeArgs.last).summonLambda[SemigroupalK](f), "productK")
+          .appliedToTypes(List(F, G))
+          .appliedTo(argf, argg)
+    }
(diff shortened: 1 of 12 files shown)


tests/src/test/scala-3/cats/tagless/tests/autoSemigroupalKTests.scala
```diff
@@ -0,0 +1,64 @@
...
+    def a[T](a: T): F[Unit]
+
+  trait algWithCurryMethod[F[_]] derives SemigroupalK:
+    def a(b: String)(d: Int): F[Unit]
+
+  trait AlgWithVarArgsParameter[F[_]] derives SemigroupalK:
+    def sum(xs: Int*): Int
+    def effectfulSum(xs: Int*): F[Int]
+
+  trait AlgWithConstantReturnTypes[F[_]] /*derives SemigroupalK*/:
```
pomadchin, line 61: This makes sense, I never added the 'simple' semigroup case into the macro :/

core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
```diff
@@ -0,0 +1,54 @@
...
+    val f = F.typeSymbol
+    val h = H.typeSymbol
+
+    List(eaf.asTerm, eag.asTerm).transform[Alg[Tuple2K[F, G, *]]] {
+      case (tpe, argf :: argg :: Nil) if tpe.contains(h) =>
+        Select
+          .unique(tpe.typeArgs.head.appliedTo(tpe.typeArgs.last).summonLambda[SemigroupalK](f), "productK")
+          .appliedToTypes(List(F, G))
+          .appliedTo(argf, argg)
+    }
```
joroKr21, line 54: Hmm ok, I see the current helpers don't work great when we have two targets.
So this is incomplete but I think we can merge it and iterate on top.

tests/src/test/scala-3/cats/tagless/tests/autoInvariantKTests.scala
```diff
@@ -0,0 +1,201 @@
...
+    def a(i: Int): Int
+
+  // trait ContravariantEff[F[_]] derives InvariantK {
+  //   def a(i: F[Int], j: String): Int
+  //   def b(i: F[Int]): Int
+  // }
+
+  // trait InvariantEff[F[_]] derives InvariantK {
+  //   def a(i: F[Int], j: String): F[Int]
+  // }
```
joroKr21, line 146: Those don't work?

joroKr21 (commented):
Looks good, but to me it seems like you could uncomment more of the `InvariantK` tests.

tests/src/test/scala-3/cats/tagless/tests/autoInvariantKTests.scala
pomadchin: Uncommented: https://github.com/typelevel/cats-tagless/pull/513/commits/7ca7c66cd91debf4deda4cca66b90d00910e7bef

pomadchin:
I think I can improve SemigroupalK a bit; but mb it can be an iterative followup PR; I oversimplified the SemigroupalK macro - we also need do to smth with F[_] args there + the semigroup case support.

core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
pomadchin: I tried to keep your style; mb we can have a single generic extension with the `transform` method for the `Seq[Term]` 🤔

joroKr21: The complication is how to transform the arguments of `eaf` in one way and of `eag` in another way.

tests/src/test/scala-3/cats/tagless/tests/autoInvariantKTests.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  trait AlgWithVarArgsParameter[F[_]] derives InvariantK:
+    def sum(xs: Int*): Int
+    def covariantSum(xs: Int*): F[Int]
+    // def contravariantSum(xs: F[Int]*): Int
+    // def invariantSum(xs: F[Int]*): F[Int]
+
+  // trait AlgWithByNameParameter[F[_]] derives InvariantK:
+  //   def whenM(cond: F[Boolean])(action: => F[Unit]): F[Unit]
+  //
```
joroKr21, line 188: Those don't work?

pomadchin: `AlgWithByNameParameter` does not work because of a by -name parameter (didn't have a chance to look closer): 

```scala
no implicit values were found that match type cats.tagless.InvariantK[[G[_$6]] =>> => G[Unit]]
```

and `AlgWithVarArgsParameter` due to varargs (invariantK is not that smart atm):

```scala
no implicit values were found that match type cats.tagless.InvariantK[[G[_$6]] =>> G[Int]*]
```

joroKr21: Cool, we can look at those later

joroKr21 (approved):

PR #520, opened 2024-04-01
Hacky solution of the Tuple2K problem

joroKr21:
Substitute `Tuple2K` with a type alias that picks the first type.

### DIFF
--- a/core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
@@ -38,39 +38,44 @@ object MacroSemigroupalK:
   )(using q: Quotes): Expr[Alg[Tuple2K[F, G, *]]] =
     import quotes.reflect.*
     given DeriveMacros[q.type] = new DeriveMacros
-    extension (tpe: TypeRepr) def fromTuple2K: TypeRepr = tpe.typeArgs.head.appliedTo(tpe.typeArgs.last)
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala-3/cats/tagless/macros/macroSemigroupalK.scala
```diff
@@ -38,39 +38,44 @@ object MacroSemigroupalK:
...
+    val T2K = TypeRepr.of[Tuple2K[F, G, *]]
     val f = F.typeSymbol
-    val h = H.typeSymbol
+    val t2k = T2K.typeSymbol
 
-    def tuple2K(name: String): Term = Select.unique('{ SemigroupalK }.asTerm, name).appliedToTypes(List(F, G))
+    type FirstK[F[_], G[_], A] = F[A]
+    extension (tpe: TypeRepr)
+      def firstK: TypeRepr =
+        tpe.substituteTypes(t2k :: Nil, TypeRepr.of[FirstK] :: Nil)
```
pomadchin, line 51: haha, `substituteTypes` is the thing I could not find :D

joroKr21: Using the original `F` type is not as simple as it sounded 😢 
It also wasn't easy to come up with this hack 😄

pomadchin: lol scala macros stuff is just its own language

PR #523, opened 2024-04-05
Bump tlBaseVersion

joroKr21:


bpholt:
Was this necessary? I didn't see any build failures from MiMa indicating a break in bincompat.

joroKr21:
> Was this necessary? I didn't see any build failures from MiMa indicating a break in bincompat.

We are adding new classes and methods. Perhaps we could do that in a patch release since the major version is still zero 🤔 
But previously releases also bumped the minor version on major additions so at least it's consistent.

bpholt:
It's unfortunate though, because this change indicates a major version bump in early semver, which means anything downstream should also do a major version bump. This causes issues with the eviction process when libraries depend on cats-tagless.

This was the reason I added #430 last year, because there was a version bump when there shouldn't have been at that time too. That PR added the bincompat checks so downstream users can safely ignore eviction warnings when moving from 0.14→0.15. Now we should probably add the same thing for 0.16. (It might be too late to release 0.15.1 and tell Scala Steward not to make PRs for 0.16.)

Don't get me wrong, I'm very excited about the recent developments and this release. It's just that since the version was changed, it makes it harder downstream to pull in those changes.

joroKr21:
> This was the reason I added https://github.com/typelevel/cats-tagless/pull/430 last year, because there was a version bump when there shouldn't have been at that time too. That PR added the bincompat checks so downstream users can safely ignore eviction warnings when moving from 0.14→0.15. Now we should probably add the same thing for 0.16. (It might be too late to release 0.15.1 and tell Scala Steward not to make PRs for 0.16.)

True, I totally forgot about that 🤔 - what about just changing to SemVer even though it's 0? We haven't broken binary compatibility for a long time, I don't think anyone should be using the old versions.

joroKr21:
@bpholt do you depend on cats-tagless in an OSS project? I would be happy to try it out.

bpholt:
This is the main one: https://github.com/dwolla/async-utils

So are you suggesting adding something to the published artifact to tell sbt to use semver instead of early-semver? I'm not 100% sure how that works, but that might work.

joroKr21:
> So are you suggesting adding something to the published artifact to tell sbt to use semver instead of early-semver? I'm not 100% sure how that works, but that might work.

No, I'm suggesting to release a new artifact with semver but I'm not sure if it will work either.

bpholt:
I think [`versionScheme`](https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme) is the key to set to indicate what the project uses, but I'm not sure any of the supported options are good for our use case here. 🤔

It's currently `early-semver` (as the default). I don't _think_ `semver-spec` does what we want, because if I understand it correctly, it would indicate that `0.16.0` and `0.16.1` are not necessarily binary compatible, let alone `0.15.0` and `0.16.0`. 

In any event, it looks like sbt-typelevel only supports `early-semver`. I tried setting it locally to `semver-spec` and got an
```
java.lang.IllegalArgumentException: requirement failed: Only early-semver versioning scheme supported.
	at scala.Predef$.require(Predef.scala:281)
	at org.typelevel.sbt.TypelevelMimaPlugin$.$anonfun$buildSettings$2(TypelevelMimaPlugin.scala:48)
… snip …
```
when loading the build.

Maybe @armanbilge has some suggestions for us?

I guess if we're feeling good about the bincompat story, we could release `1.0.0` and go from there?

armanbilge:
> In any event, it looks like sbt-typelevel only supports `early-semver`

That's right.

> I don't _think_ `semver-spec` does what we want, because if I understand it correctly, it would indicate that `0.16.0` and `0.16.1` are not necessarily binary compatible

This is also my understanding. It basically means no compatibility guarantees at all until 1.x.

> I guess if we're feeling good about the bincompat story, we could release `1.0.0` and go from there?

Yes. You can do like cats-parse did recently and continue checking bincompat against the last `0.x` series.

joroKr21:
I'm not sure releasing 1.0 is the best idea since all the Scala 3 macros are still experimental 😢

bpholt:
That makes sense! In the meantime, I would request that we not bump the minor version unless forced to by MiMa, and release new features as patch releases off `0.16` going forward.

joroKr21:
> That makes sense! In the meantime, I would request that we not bump the minor version unless forced to by MiMa, and release new features as patch releases off 0.16 going forward.

I'm gonna add a comment to #524 - should've done that last time 👍

PR #534, opened 2024-04-29
[Scala 3] Derive Bifunctor

joroKr21:
@pomadchin it seems to work fine

### DIFF
--- a/core/src/main/scala-3/cats/tagless/macros/MacroBifunctor.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/MacroBifunctor.scala
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2019 cats-tagless maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.tagless.macros
+
+import cats.tagless.*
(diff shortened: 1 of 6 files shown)


core/src/main/scala-3/cats/tagless/macros/MacroBifunctor.scala
```diff
@@ -0,0 +1,87 @@
...
+            .unique(tpe.summonLambda[Functor](c), "map")
+            .appliedToTypes(List(A, C))
+            .appliedTo(body)
+            .appliedTo(f.asTerm)
+        case (tpe, body) if tpe.contains(d) =>
+          Select
+            .unique(tpe.summonLambda[Functor](d), "map")
+            .appliedToTypes(List(B, D))
+            .appliedTo(body)
+            .appliedTo(g.asTerm)
```
pomadchin, line 85: haha I see whtat you've done; I stopped yesterday at 

```scala
case (tpe, body) if tpe.contains(c) =>
          Select
            .unique(tpe.summonLambda[Functor](c), "map")
            .appliedToTypes(List(A, C))
            .appliedTo(body)
            .appliedTo(f.asTerm)
        case (tpe, body) if tpe.contains(d) =>
          Select
            .unique(tpe.summonLambda[Functor](d), "map")
            .appliedToTypes(List(B, D))
            .appliedTo(body)
            .appliedTo(g.asTerm)
```

core/src/main/scala-3/cats/tagless/macros/MacroBifunctor.scala
```diff
@@ -0,0 +1,87 @@
...
+    val B = TypeRepr.of[B]
+    val C = TypeRepr.of[C]
+    val D = TypeRepr.of[D]
+    val c = C.typeSymbol
+    val d = D.typeSymbol
+
+    fab.asTerm.transformTo[F[C, D]](
+      args = {
+        case (tpe, arg) if tpe.contains(c) && tpe.contains(d) =>
+          report.errorAndAbort(s"Both type parameters ${A.show} and ${B.show} appear in contravariant position")
```
pomadchin, line 53: This is a cool catch

joroKr21: There is no `Bicontravariant` 😄

PR #554, opened 2024-06-20
Add macro for autoderiving Instrument in Scala 3

alonsodomin:
A (naive) stab at adding a new Scala 3 macro for auto deriving the `Instrument` type class.


core/src/main/scala-3/cats/tagless/macros/MacroInstrument.scala
```diff
@@ -0,0 +1,61 @@
...
+  )(using q: Quotes): Expr[Alg[[X] =>> Instrumentation[F, X]]] =
+    import quotes.reflect.*
+    given DeriveMacros[q.type] = new DeriveMacros
+
+    val F = TypeRepr.of[F]
+    val Alg = TypeRepr.of[Alg]
+
+    alg.transformTo[Alg[[X] =>> Instrumentation[F, X]]](
+      body = {
+        case (sym, tpe, body) if tpe.contains(F) =>
```
joroKr21, line 49: We should check `tpe <:< TypeRepr.of[Instrumentation[F, ?]]` instead

joroKr21 (commented):
Otherwise LGTM 👍

joroKr21 (approved):
Nice, thanks!

PR #556, opened 2024-06-26
[Scala 3] Autoderivation for the Aspect type class

alonsodomin:
Attempt at implementing the auto derivation macro for the `Aspect` type class.

Submitting this more as a way of getting other pair of eyes to look over it than as a final implementation, someone that knows about Scala 3 macros more than me. It currently doesn't compile the `AspectTests.ShowFAlgebra` when deriving for the `Show` type class (more info in the line comment)

This implementation is based on quotes but also tried building the tree manually without success.


core/src/main/scala-3/cats/tagless/macros/MacroAspect.scala
```diff
@@ -0,0 +1,116 @@
...
+            val isByName = param.isByName
+            val paramValue = Ref(param.symbol)
+
+            param.tpt.tpe.widenParam.asType match
+              case '[t] if isByName =>
+                '{
+                  Aspect.Advice.byName[Dom, t](
+                    name = ${ Expr(param.name) },
+                    thunk = ${ paramValue.asExprOf[t] }
+                  )(using summonInline)
```
alonsodomin, line 73: The compilation error comes from here, there generated code looks like the following:

```scala
override def logF[A](message:  => A)(implicit evidence$1: cats.Show[A]): cats.tagless.aop.Aspect.Weave[F, [T >: scala.Nothing <: scala.Any] => cats.Show[T], [T >: scala.Nothing <: scala.Any] => cats.Show[T], scala.Unit] =
  cats.tagless.aop.Aspect.Weave.apply[F, [T >: scala.Nothing <: scala.Any] => cats.Show[T], [T >: scala.Nothing <: scala.Any] => cats.Show[T], scala.Unit](
    algebraName = "ShowFAlgebra",
    domain = scala.List.apply[scala.collection.immutable.List[cats.tagless.aop.Aspect.Advice[[A >: scala.Nothing <: scala.Any] => cats.Eval[A], [T >: scala.Nothing <: scala.Any] => cats.Show[T]]]](
      scala.List.apply[cats.tagless.aop.Aspect.Advice[[A >: scala.Nothing <: scala.Any] => cats.Eval[A], [T >: scala.Nothing <: scala.Any] => cats.Show[T]]](
        cats.tagless.aop.Aspect.Advice.byName[[T >: scala.Nothing <: scala.Any] => cats.Show[T], A](
          name = "message",
          thunk = message
        )(scala.compiletime.package$package.summonInline[cats.Show[A]])
      )
    ),
    codomain = cats.tagless.aop.Aspect.Advice.apply[F, [T >: scala.Nothing <: scala.Any] => cats.Show[T], scala.Unit](
      adviceName = "logF",
      adviceTarget = af.logF[A](message)(evidence$1)
    )(scala.compiletime.package$package.summonInline[cats.Show[scala.Unit]])
  )
```

The offending bit is `scala.compiletime.package$package.summonInline[cats.Show[A]]` even though the method declares an implicit parameter `Show[A]`. My understanding is that `summonInline` would take into account that implicit parameter, however it looks like it's being inlined within the implicit scope of the method itself.

Using trees instead of quotes to achieve the same makes it fail earlier though. Calling `byName` (or `byValue`) without explicitly passing the summoned implicit requires for eta expansion of the `byName` constructor and when building the list of parameter lists for the codomain, the types do not align anymore, so the code generation itself fails (instead of the compilation of the generated code as in the current case).

In the other hand, summoning the missing implicit using trees fails too as the implicit scope at that point doesn't know anything about `Show[A]`.

There is a workaround for this specific case: look for the required implicit from the implicit parameter list of the given method and pass that one if found or fallback into `summonInline`. However that also fails for other cases like varargs or those in which the required implicit must be derived on the fly (i.e. `Show[(A, String)]`).

Have the feeling that I'm just missing a small detail and there is an easy solution for this, but after days printing trees to stdout I would appreciate someone could point it out.

joroKr21: That is a difficult one. It may simply come down to the order of inlining / macro expansion in Scala 3. Still thinking if we can do anything about it. Inspecting the parameters by hand sounds quite hacky.

joroKr21: Maybe we just accept it won't work for now and drop this test?

alonsodomin: yeah, makes sense to try to move on with a limited solution.

alonsodomin:
Added a macro for deriving algebras based on `Const` or `Void`, which allows enabling one of the aspect related tests.

pomadchin:
@alonsodomin btw I added const / void yesterday along with some extra macro syntax in a separate PR: https://github.com/typelevel/cats-tagless/pull/560 mb we can keep PRs separate for now haha

pomadchin:
I wonder if it makes sense to change PRs merge order

alonsodomin:
@pomadchin oops! missed that, thanks for pointing it out, I'll revert the last two commits then and wait for that other one to merge.

ps: cool refactoring for the `ReaderT` macro btw.

joroKr21 (approved):

joroKr21:
Nice, I simplified it a bit 👍

PR #557, opened 2024-06-27
Auto derive Scala 3 macro for ReaderT

alonsodomin:
Initial implementation of a macro for auto-derivation of algebras based on `ReaderT`, which results useful to transform side-effecting algebra implementations into algebras using cats-effect `IO`.

The implementation holds some duplication in relation to the common code found in `DeriveMacros` utility extensions. It's been left in the macro implementation because, even though there are some commonalities, the derivation of `ReaderT` has its own peculiarities that may not be worth a refactor on the common utilities. Up to discussion, obviously.

### DIFF
--- a/core/src/main/scala-3/cats/tagless/Derive.scala
+++ b/core/src/main/scala-3/cats/tagless/Derive.scala
@@ -17,6 +17,7 @@
 package cats.tagless
 
 import cats.*
+import cats.data.ReaderT
 import cats.arrow.Profunctor
 import cats.tagless.*
 import cats.tagless.aop.*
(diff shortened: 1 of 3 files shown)


joroKr21 (approved):
Thank you very much @alonsodomin!

I hope you don't mind, I pushed some changes to your branch instead of reviewing.
I wasn't sure what I had in mind would actually work until I tried it.

kubukoz:
Very exciting! Can we have a patch with this? 🥺

joroKr21:
Yes, I wanted to merge https://github.com/typelevel/cats-tagless/pull/556 first though

PR #560, opened 2024-07-11
[Scala 3] Add derivation for Const and Void

pomadchin:
This PR adds a `ConstMacro` and both `Derive.const` / `Derive.void` functions. 
I also added an extra helper function, apparently it's used a lot!

Connected https://github.com/typelevel/cats-tagless/issues/170

### DIFF
--- a/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
@@ -27,9 +27,9 @@ object MacroConst:
 
   private[macros] def deriveConst[Alg[_[_]]: Type, A: Type](const: Expr[A])(using q: Quotes): Expr[Alg[Const[A]#λ]] =
     import quotes.reflect.*
-    given dm: DeriveMacros[q.type] = new DeriveMacros
+    given DeriveMacros[q.type] = new DeriveMacros
 
-    dm.newClassOf[Alg[Const[A]#λ]](
+    Symbol.newClassOf[Alg[Const[A]#λ]](
       transformDef = _ => _ => Some(const.asTerm),
       transformVal = _ => Some(const.asTerm)
     )
--- a/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2019 cats-tagless maintainers
+ *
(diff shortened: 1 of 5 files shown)


alonsodomin (approved):

core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
```diff
@@ -0,0 +1,35 @@
...
+  inline def derive[Alg[_[_]], A](const: A): Alg[Const[A]#λ] = ${ deriveConst[Alg, A]('const) }
+
+  private[macros] def deriveConst[Alg[_[_]]: Type, A: Type](const: Expr[A])(using q: Quotes): Expr[Alg[Const[A]#λ]] =
+    import quotes.reflect.*
+    given DeriveMacros[q.type] = new DeriveMacros
+
+    Symbol.newClassOf[Alg[Const[A]#λ]](
+      transformDef = _ => _ => Some(const.asTerm),
+      transformVal = _ => Some(const.asTerm)
+    )
```
joroKr21, line 35: We need to limit the transformations to those that actually return `A`

tests/src/test/scala/cats/tagless/tests/ConstTests.scala
joroKr21, line 1: Did you just move this test? We need to keep it in both places.

core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
pomadchin: ha, how does it work in the Scala2 macro now? Does it overwrite the return types of all methods?

pomadchin: `method.copy(returnType = weakTypeOf[A], body = q"$tmp")` :o 

should we do the same or we should forbid const generation for anything that does not match the signature?

joroKr21: Ah sorry - my bad. We have no other receiver to delegate methods that don't return `A` so those will simply fail to compile. It's good like this 👍

joroKr21 (approved):

core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
pomadchin: LOL that Const usage is really tricky the mind ha :D

PR #561, opened 2024-07-11
Replace FunctionKLift with FunctionK.liftFunction

pomadchin:
This PR removes FunctionKLift and replaces it via [FunctionK.liftFunction](https://github.com/typelevel/cats/pull/4603)


joroKr21 (approved):
Looks like we need MiMa filters although it was `private[tagless]`

PR #562, opened 2024-07-12
[Scala 3] MacroConst and MacroReaderT Errors readability improvement

pomadchin:
I still took a closer look into the comment left here https://github.com/typelevel/cats-tagless/pull/560#discussion_r1675658431 

It is true that macro generated code fails to compile & errors produced are OK but a bit cryptic:

```scala
[error] 31 |    val safe = Derive.const[NotSimpleAlg, Int](42)
[error]    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |Malformed tree was found while expanding macro with -Xcheck-macros.
[error]    |               |The tree does not conform to the compiler's tree invariants.
[error]    |               |
[error]    |               |Macro was:
[error]    |               |scala.quoted.runtime.Expr.splice[ConstTests.this.NotSimpleAlg[cats.tagless.Const[scala.Int].<none>]](((evidence$1: scala.quoted.Quotes) ?=> cats.tagless.macros.MacroConst.inline$deriveConst[[F >: scala.Nothing <: [_$1 >: scala.Nothing <: scala.Any] => scala.Any] => ConstTests.this.NotSimpleAlg[F], scala.Int](scala.quoted.runtime.Expr.quote[scala.Int](42).apply(using evidence$1))(scala.quoted.Type.of[[F >: scala.Nothing <: [_$1 >: scala.Nothing <: scala.Any] => scala.Any] => ConstTests.this.NotSimpleAlg[F]](evidence$1), scala.quoted.Type.of[scala.Int](evidence$1), evidence$1)))
[error]    |               |
[error]    |               |The macro returned:
[error]    |               |{
[error]    |  class $anon$macro$1 extends java.lang.Object with ConstTests.this.NotSimpleAlg[[T >: scala.Nothing <: scala.Any] => scala.Int] {
[error]    |    override def str(str: scala.Predef.String): scala.Predef.String = 42
[error]    |    override def parseInt(`str₂`: scala.Predef.String): scala.Int = 42
[error]    |    override def divide(dividend: scala.Float, divisor: scala.Float): scala.Int = 42
[error]    |  }
[error]    |  new $anon$macro$1()
[error]    |}
[error]    |               |
[error]    |               |Error:
[error]    |               |assertion failed: Found:    (42 : Int)
[error]    |Required: String
[error]    |found: ??
[error]    |expected: type String in object Predef with type String, flags = <touched>, underlying = String,  = String, String, Object with Serializable with Comparable[String] with CharSequence {...}
[error]    |tree = 42
[error]    |               |
[error]    |stacktrace available when compiling with `-Ydebug`
[error]    |               |
[error]    |----------------------------------------------------------------------------
[error]    |Inline stack trace
```

This is an attempt to align it a bit with the Scala 2 Macro in terms of error handling:

```scala
[error] 44 |    Derive.const[NotSimpleAlg, Int](42)
[error]    |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |   Expected method str to return scala.Int but found scala.Predef.String
[error]    |----------------------------------------------------------------------------
[error]    |Inline stack trace
[error]    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]    |This location contains code that was inlined from ConstTests.scala:44
[error] 45 |  inline def const[Alg[_[_]], A](value: A): Alg[Const[A]#λ] = MacroConst.derive[Alg, A](value)
[error]    |                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     ----------------------------------------------------------------------------
[error] one error found
```

### DIFF
--- a/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
@@ -29,7 +29,24 @@ object MacroConst:
     import quotes.reflect.*
     given DeriveMacros[q.type] = new DeriveMacros
 
-    Symbol.newClassOf[Alg[Const[A]#λ]](
-      transformDef = _ => _ => Some(const.asTerm),
-      transformVal = _ => Some(const.asTerm)
-    )
+    val A = TypeRepr.of[A]
+
+    def transformDef(method: DefDef)(argss: List[List[Tree]]): Option[Term] =
+      method.returnTpt.tpe.asType match
+        case '[A] =>
+          Some(const.asTerm)
+        case _ =>
+          report.errorAndAbort(
+            s"Expected method ${method.name} to return ${A.show} but found ${method.returnTpt.tpe.show}"
+          )
+
+    def transformVal(value: ValDef): Option[Term] =
+      value.tpt.tpe.asType match
+        case '[A] =>
+          Some(const.asTerm)
+        case _ =>
+          report.errorAndAbort(
+            s"Expected value ${value.name} to be of type ${A.show} but found ${value.tpt.tpe.show}"
+          )
+
+    Symbol.newClassOf[Alg[Const[A]#λ]](transformDef, transformVal)
--- a/core/src/main/scala-3/cats/tagless/macros/MacroReaderT.scala
+++ b/core/src/main/scala-3/cats/tagless/macros/MacroReaderT.scala
@@ -51,13 +51,17 @@ object MacroReaderT:
               yield argTransformer(af).transformArg(method.symbol, paramAndArg)
           )
         case _ =>
-          method.rhs
+          report.errorAndAbort(
+            s"Expected method ${method.name} to return ${RT.show} but found ${method.returnTpt.tpe.show}"
+          )
 
     def transformVal(value: ValDef): Option[Term] =
       value.tpt.tpe.asType match
         case '[ReaderT[F, Alg[F], t]] =>
           Some(readerT[t](value.symbol)(_.select(value.symbol)))
         case _ =>
-          value.rhs
+          report.errorAndAbort(
+            s"Expected value ${value.name} to be of type ${RT.show} but found ${value.tpt.tpe.show}"
+          )
 
     Symbol.newClassOf[Alg[[X] =>> ReaderT[F, Alg[F], X]]](transformDef, transformVal)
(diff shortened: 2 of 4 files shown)


alonsodomin:
same thing could also be added to the `MacroReaderT`. At the moment the implementation fallsback into returning `method.rhs` or `value.rhs` if the expression type doesn't match the expected one to be transformed.

But since `MacroReaderT` creates a fresh implementation from scratch without possibly delegating execution (as `MacroConst` does), the `rhs` value in both cases will be `None` and therefore ending up in a cryptic error message.

pomadchin:
@alonsodomin ha, you're right! I can add it there as well 🤔

pomadchin:
The Scala2 version also catches it to print over there https://github.com/typelevel/cats-tagless/blob/master/macros/src/main/scala-2/cats/tagless/DeriveMacros.scala#L746

pomadchin:
@alonsodomin actually in case of ReaderT macro it gonna print "Concrete method has no definition: override def $methodName: String", I think it makes sense. I can modify it to be more specific though, i.e. 'why' the macros function failed.

pomadchin:
Added ReaderT error handling here => 4d66cd6

alonsodomin:
isn't it though the same case as with `Const`?

I mean, the macro is building an instance of `Alg[F]` where `F` is `[X] =>> ReaderT[F, Alg[F], X]`, so the case in which the return type of any method or value in `Alg` not be `ReaderT[F, Alg[F], t]` should be because in `trait Alg[F[_]]` the given method or value was not defined to return `F[t]`.

In other words, the case in which this match would fail:

```scala
value.tpt.tpe.asType match
        case '[ReaderT[F, Alg[F], t]] =>
          Some(readerT[t](value.symbol)(_.select(value.symbol)))
```

would be because the type of `value` is not `F[t]`, the usage of `case '[ReaderT[F, Alg[F], t]]` is more to make it handy to extract the type `t` so it can be used to create the instance of `ReaderT` ...

Or am I reading it incorrectly?

pomadchin:
So for the Const case Macro was building incorrect Trees (that caused these cryptic errors), smth like:
```scala
def str: String = 42
```

The reason ie because we were [unconditionally appending the same function body to any function signature](https://github.com/typelevel/cats-tagless/blob/master/core/src/main/scala-3/cats/tagless/macros/MacroConst.scala#L33-L34)

With ReaderT it's been left as unimplemented, that's why the error is a bit better.

But regardless it feels like a good idea to handle it.

joroKr21 (approved):
Better error handling is definitely a good thing 👍

joroKr21:
I see what you mean. The difference between Scala 2 and Scala 3 is that in Scala 2 we work with the original type of the algebra whereas in Scala 3 we work with the target type. I think it would be fine to adjust the error message.

PR #577, opened 2024-10-27
Bump Scala to 3.9, drop hacks

joroKr21:
Use `-experimantal` compiler option


core/src/main/scala-3/cats/tagless/macros/DeriveMacros.scala
```diff
@@ -138,24 +138,40 @@ private class DeriveMacros[Q <: Quotes](using val q: Q):
...
+        member <- sym.typeMembers
+        if member.isTypeDef
+        tpe = delegate match
+          case Some(delegate) => TypeSelect(delegate, member.name).tpe
+          case None => report.errorAndAbort(s"Not supported: $member in $sym")
+      yield member -> tpe
+
+      val cls = This(sym).tpe
+      val aliases = typeAliases.toMap
+      val (from, to) = typeAliases.unzip
```
pomadchin, line 153: Ha just curious, how exactly the type members support is implemented now?

build.sbt
```diff
@@ -8,7 +8,7 @@ addCommandAlias("fmtCheck", "all scalafmtSbtCheck scalafmtCheckAll")
 
 val Scala212 = "2.12.20"
 val Scala213 = "2.13.15"
-val Scala3 = "3.3.4"
+val Scala3 = "3.6.1"
```
bpholt, line 11: Is it possible to split this into a separate module that can be compiled with 3.6, leaving the existing modules on the LTS version? That will avoid other libraries that depend on cats-tagless from having to update beyond LTS unless they want to use this functionality.

core/src/main/scala-3/cats/tagless/macros/DeriveMacros.scala
joroKr21: Curious about what? Sorry, I didn't get the question.

build.sbt
joroKr21: Unfortunately that's at the core of the library. I don't plan to merge this anytime soon but it's a good question. Do you think it's possible to somehow achieve this?

core/src/main/scala-3/cats/tagless/macros/DeriveMacros.scala
pomadchin: Oh yesterday night I checked the code and didn’t get where type members are coming from haha; all good now ic it :D

joroKr21: Surprisingly complicated, I thought it would be enough to define the type member in the target class, but it looks like other members' types aren't automatically substituted.

build.sbt
bpholt: Good question. I saw the PR come through on my phone initially and didn't have a chance to look at it in more detail before I asked. Looking at it now, I don't immediately have any good ideas. 🙁 

I wonder when the next LTS is due to be released. If it's two years from the release of v3.3.0, then I suppose it would be May 2025? It would be unfortunate to have to wait that long.

I'll keep thinking about it in the meantime. 🤔

joroKr21: We need some custom cross-building suffix I guess. I will ask on Discord when I get some free time.

build.sbt
```diff
@@ -141,7 +141,7 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(rootSettings, macroSettings)
   .settings(
     moduleName := "cats-tagless-tests",
-    scalacOptions ++= when(scalaBinaryVersion.value == "3")("-Xcheck-macros"),
+    scalacOptions ++= when(scalaBinaryVersion.value == "3")("-experimental", "-Xcheck-macros"),
```
pomadchin, line 144: :+1:

core/src/main/scala-3/cats/tagless/macros/DeriveMacros.scala
```diff
@@ -40,26 +40,6 @@ private class DeriveMacros[Q <: Quotes](using val q: Q):
...
-      val decorators = decoratorsClass.getDeclaredField("MODULE$").get(null)
-      val toTypeName = decorators.getClass.getMethods.find(_.getName == "toTypeName").get
-      val newSymbol = symbols.getClass.getMethods.find(_.getName == "newSymbol").get
-      val typeName = toTypeName.invoke(decorators, name)
-      val nestingLevel = ctx.getClass.getMethod("nestingLevel").invoke(ctx)
-      val sym = newSymbol.invoke(symbols, ctx, owner, typeName, flags, alias, privateIn, 0, nestingLevel)
-      sym.asInstanceOf[Symbol]
-    catch
-      case _: Exception =>
-        report.errorAndAbort(s"Not supported: type $name in $owner")
```
pomadchin, line 61: TIL

joroKr21: I got tired of waiting 😄 - if we have to live with dirty hacks, at least they will be in library code

build.sbt
```diff
@@ -151,7 +152,7 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(rootSettings, macroSettings)
   .settings(
     moduleName := "cats-tagless-tests",
-    scalacOptions ++= when(tlIsScala3.value)("-Xcheck-macros"),
+    scalacOptions ++= when(tlIsScala3.value)("-experimental", "-Xcheck-macros"),
```
kubukoz, line 155: do we still need `-experimental`? I thought the whole thing with 3.9 was that we don't need it anymore

core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
```diff
@@ -20,6 +20,7 @@ import cats.tagless.*
 
 import scala.annotation.*
 import scala.quoted.*
+import scala.annotation.unused
 
 @experimental
```
kubukoz, line 25: is this a leftover?

build.sbt
joroKr21: Some macro APIs are still experimental 🤷

core/src/main/scala-3/cats/tagless/macros/MacroConst.scala
joroKr21: Nope: https://github.com/scala/scala3/blob/main/library/src/scala/quoted/Quotes.scala#L4841-L4843

build.sbt
kubukoz: big sigh

PR #620, opened 2025-05-09
Replace usages of changeOwner

joroKr21:


core/src/main/scala-3/cats/tagless/macros/MacroFlatMap.scala
```diff
@@ -77,13 +77,14 @@ object MacroFlatMap:
...
       body = {
         case (method, tpe, body) if tpe =:= B =>
+          given Quotes = method.asQuotes
           '{
             @tailrec def step(x: A): B =
               ${ body.replace(a, '{ x }).asExprOf[Either[A, B]] } match
                 case Left(a) => step(a)
                 case Right(b) => b
             step($a)
-          }.asTerm.changeOwner(method)
```
pomadchin, line 86: just curious, what changed in there? 👀

joroKr21: > Fix https://github.com/scala/scala3/issues/20471: owners of top-level symbols in cached quoted code being incorrect https://github.com/scala/scala3/pull/21945

PR #647, opened 2025-10-14
Add documentation for each macro annotation (issue #7)

PATIENCE-ONAH:
Updated the doc with the documentation of each macro annotation covering the following:

1. Intent of the annotation
2. Scope of the annotation
3. Code snippet showing annotated traits before and after applying macro
4. Dependencies and Limitations
5. Test Behavior

### DIFF
--- a/docs/docs/macros/autoApply.md
+++ b/docs/docs/macros/autoApply.md
@@ -1,0 +1,78 @@
+---
+layout: page
+title: @autoApply
+section: macros
+position: 5
+---
+
+
+## @autoApply
+
+### Intent
+Generates an instance of `cats.Apply` in the companion object of the annotated trait.
+
+### Scope
+The annotation can be applied to traits:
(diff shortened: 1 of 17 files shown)


armanbilge (commented):
Thanks for the PR! Since this is your first contribution, let's start by focusing on documenting just one macro annotation. I suggest `@autoFunctor`.

For the example, I suggest using a `case class` (or ADT hierarchy) as a concrete example for readers. When demonstrating the result of the macro, I think it is most helpful to show what the implementation of the typeclass looks like.

By the way, even though this is marked as a good-first-issue, I do think this is a challenging issue to work on if you are new to functional programming and the Typelevel ecosystem.

PATIENCE-ONAH:
Thank you for the review and your time.
Let me try my best and update the @autofunctor based on your remark.
Thanks again.

